PR #18, opened 2018-02-13
combine PackageCompiler with static-julia

SimonDanisch:


SimonDanisch:
after discussing the future of PackageCompiler & static-julia, @ViralBShah and I concluded it would be nice to merge the efforts and include the features from static-julia into PackageCompiler.

You may ask, why not include PackageCompiler into static-julia?
I argue that static-julia needs to be a julia package, to gracefully handle dependencies (e.g. install gcc on windows) and to make it possible for users to depend on PackageCompiler! Also like this we can use the usual CI integration.

CC: @lucatrv

examples/program.c
```diff
@@ -0,0 +1,35 @@
...
+    jl_options.image_file = JULIAC_PROGRAM_LIBNAME;
+    julia_init(JL_IMAGE_JULIA_HOME);
+
+    // Do some work
+    julia_main();
+
+    // Cleanup and graceful exit
+    jl_atexit_hook(0);
+    return 0;
+}
```
vtjnash: `example2.c` is a better example file. I think we should just delete this one.

ViralBShah:
Also pinging @keno @stefankarpinski @vtjnash for their thoughts.

I do thinking having all of this in a package is the right way to go for now to make it easy to install and make it available to package authors.

vtjnash:
:+1:

This also aligns well with my original goal of keeping `static-julia` as a small demo repo for demonstrating and experimenting with the minimum required to use the right compilation flags.

StefanKarpinski:
Seems like a good idea to me as well.

src/snooping.jl
```diff
@@ -0,0 +1,78 @@
+
+function snoop(path, compilationfile, csv)
+    cd(@__DIR__)
+    # Snoop compiler can't handle the path as a variable, so we just create a file
```
vtjnash: I thought I had fixed many of the issues here in SnoopCompile (missing `esc` in macros, need for `tmp_mod`, some cases of bad printing). Are there issues filed for these?

SimonDanisch: I will take a second look! I haven't filed any issues yet

lucatrv:
Please reconsider the name choices. I would keep the main script named `juliac.jl`, not only because people are already using it, but also because it is more meaningful. Then I would rename function `julia_compile` and its script file `juliac.jl` respectively `static_julia` and `static_julia.jl`. I was already planning this, in order to differentiate it from `julia-compile` (see: https://github.com/JuliaLang/julia/issues/15864).

ViralBShah:
Agree on retaining the name `juliac.jl`.

SimonDanisch:
Fair enough!

lucatrv:
@SimonDanisch please notice also the name of the function `julia_comile` => `static_julia` (IMHO it is better to differentiate it from [`julia-compile`](https://github.com/JuliaLang/julia/issues/15864)).
So I am going to continue developing `juliac.jl` and `static_julia.jl` here, I still have a list of things to do... with limited spare time, I should be able to complete in a couple of months, hopefully earlier.
Glad to be part of the development team! :-)

SimonDanisch:
I wanted to ship this branch with a simple working python wrapper ability, but I got into issues with linking mingw and msvc dll's for python ;) But the prototype is pretty close to completion beside that, and I will try later on linux - which should be a lot easier!
Otherwise I think I should just merge this and make the python PR separate!

lucatrv:
@ViralBShah I would add a note on [static-julia](https://github.com/JuliaComputing/static-julia) repository to refer users to [PackageCompiler.jl](https://github.com/JuliaLang/PackageCompiler.jl).

lucatrv:
@SimonDanisch after you merge this I can submit next PRs for `juliac.jl` and `static_julia.jl`.

PR #22, opened 2018-02-18
Carry -g & -O through to the `gcc` commands also.

NHDaly:
Before this commit, those flags only affect the julia parts of the
script (building object file). This commit changes juliac to also pass
`-g` and `-O` to `gcc` when building a shared library or executable.


src/static_julia.jl
```diff
@@ -122,33 +122,38 @@ function julia_compile(
         math_mode, depwarn
     )
 
-    shared && build_shared(s_file, joinpath(tmp_dir, o_file), verbose)
+    shared && build_shared(s_file, joinpath(tmp_dir, o_file), optimize, debug, verbose)
 
-    executable && compile_executable(s_file, e_file, cprog, verbose)
+    executable && compile_executable(s_file, e_file, cprog, optimize, debug, verbose)
```
NHDaly: I don't like that I had to reorder these, since the command line options are ordered `-v -O0 -g2 -a -e`, but since verbose has a [default value](#diff-d47d0c2f9ab34ed28f73bdd671edd18cL148) (`verbose = false`), it had to go at the end.

I would be happy to do any of:
 1. leave it like this,
 2. add default values to `optimize` and `debug`, or
 3. remove the default value on `verbose`.

I'd vote for 3, since there's no one even using the function anywhere without specifying that value.

SimonDanisch: I'm fine with 3!

NHDaly: Great. Changed it! Thanks!

NHDaly:
Amended & repushed!

NHDaly:
(i can squish those if you want since the first commit causes a failure. Thanks travis!)

NHDaly:
Awesome, thanks Simon!

PR #23, opened 2018-02-18
Fix line endings

lucatrv:


SimonDanisch:
Oh what? I thought I made extra sure to commit unix line endings... Thank you!

PR #36, opened 2018-02-21
Modify default program.c to also pass argv[0].

NHDaly:
Currently the default program.c is only passing through to `julia_main` the `c` `argv` arguments from index 1 through the end, and skipping `argv[0]`.

This commit changes the default program.c to pass all the argv args to julia_main, including `argv[0]`. It can be useful to know how your code was invoked.

-------------------------

Do you know if there was a reason that it was done that way? It looks like it was just copied over from the import from 2 years ago, so I assume it's okay to change? :)

What do you think? @lucatrv?
@vtjnash -- I think this was your code originally. Do you think it's okay to change the default code to pass all the argv args to julia?

### DIFF
--- a/examples/program.c
+++ b/examples/program.c
@@ -31,10 +31,10 @@ int main(int argc, char *argv[])
     // build arguments array: `String[ unsafe_string(argv[i]) for i in 1:argc ]`
     jl_array_t *ARGS = jl_alloc_array_1d(jl_apply_array_type(jl_string_type, 1), 0);
     JL_GC_PUSH1(&ARGS);
-    jl_array_grow_end(ARGS, argc - 1);
-    for (i = 1; i < argc; i++) {
+    jl_array_grow_end(ARGS, argc);
+    for (i = 0; i < argc; i++) {
         jl_value_t *s = (jl_value_t*)jl_cstr_to_string(argv[i]);
-        jl_arrayset(ARGS, s, i - 1);
+        jl_arrayset(ARGS, s, i);
     }
     // call the work function, and get back a value
     retcode = julia_main(ARGS);


NHDaly:
Here's the blame on that section:
https://github.com/JuliaLang/PackageCompiler.jl/blame/0ed06a020e4b53cf68f05997de1ded6c55ab890b/examples/program2.c#L52

It looks like it came all the way from here: https://github.com/JuliaLang/PackageCompiler.jl/commit/6ebd15bb3b0d0318215272ec815a68e674ebd50c, and so I assume there was some reason specific to the use-case of that castech_legacy program.

NHDaly:
--- ohhhhhhhh I see.... it's done that way in order to match the default Julia `ARGS` variable, which I didn't even know was a thing until just now. Is there any reason why `ARGS` doesn't give you `argv[0]`? Or, another question, is there some other way to get at that information?

I can imagine lots of reasons you might want that information, such as for example in printing a help/usage string. But in my case, I wanted to see whether my binary was being executed as part of a macOS `.app` bundle, because if so, you need to changedir to a different directory to access resources, since in mac app structure the executable sits all by itself in `MyProgram.app/Contents/MacOS/MyProgram`.

In either case, I guess we need to decide whether this wrapper `program.c` should cause the compiled program to behave more like a default `julia` program or more like a default `c` program. ... I would guess probably behaving more like a default `julia` program is preferable, in which case I should revert this PR... what do you all think?

lucatrv:
I confirm, it is to match the default Julia `ARGS` variable. Consider for instance function `julia_main` in script `juliac.jl`, which we want to run either as a script or compiled (in future `juliac.jl` will autocompile). So in my opinion it is preferable to exclude `argv[0]`.

NHDaly:
Okay, thanks for the explanation! I think that makes sense. Closing this now! :)

PR #38, opened 2018-02-21
build from vanilla backup image

### LINKED ISSUE #34: compiling a new sysimg should use vanilla sysimg as base
If I am not wrong, from the look of the precompile lists , it seems to me that
if I build a package into the sysimg and then at a later date update that package and therefore would like to build the package again into the sysimg.. well since the previous sysimg gets loaded into the new snooping process , the end result is a kind of cumulative compilation .. which is not what I intended

SimonDanisch:
should fix #34

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ compile_package("Matcha", "relative/path/for_snooping.jl")
 revert()
 
 # Or if you simply want to get a native system image e.g. when you have downloaded the generic Julia install:
-build_native_image()
+force_native_image!()
 
 # building an executable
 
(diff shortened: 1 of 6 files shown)


lucatrv:
@SimonDanisch I wonder why you removed the `tmp_dir = "tmp_v$VERSION"` directory with commit "build from vanilla backup image". Now `static_julia` saves all the cached module image files `.ji` and the object file `.o` in the main build directory. Those files (in particular the module image files) are temporary and do not need to be delivered with the main executable for it to run. Moreover the cached `.ji` files are valid only for the specific Julia version used to build them, so the directory was named with "$VERSION" to allow using different Julia versions at the same time. If you agree I can restore the `tmp` directory in my next PR.

lucatrv:
@SimonDanisch I think I found a good way to bring the `tmp` directory back while keeping the `build_object` function general, please hold on and see if you agree with my next PR.

PR #41, opened 2018-03-03
Add --ccflags option to allow custom compiler flags.

NHDaly:
Allows the user to specify compiler flags to be passed to `cc` for
building the shared object and the final executable.

---------

@lucatrv: I'm not sure about the exact name of this flag. Maybe it should just
be `cflags`? Or maybe `extra_cflags` or something?

Thanks!

### DIFF
--- a/juliac.jl
+++ b/juliac.jl
@@ -99,8 +99,8 @@ Base.@ccallable function julia_main(args::Vector{String})::Cint
         "--ccflags"
             arg_type = String
             default = nothing
-            # metavar = ""  # TODO(NHDaly): Any suggestions on what to put here?
-            help = "custom flag(s) to pass to compiler when building shared library or executable"
+            metavar = "<flags>"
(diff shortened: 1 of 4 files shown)


lucatrv:
I would consider:

```
"--cc-flags"
arg_type = String
default = nothing
metavar = "<flags>"
help = "pass additional flags to system compiler when building shared library or executable"
```
Then in `static_julia.jl` I would pass `cc_flags` to the `julia_flags` function, so it creates the full list of flags.

Thanks

lucatrv:
One reason for `--cc-flags` is because I am already adding also a `--cc` flag for custom system compiler, and I think they would list well together.

NHDaly:
Sounds good to me! I've updated to `cc-flags`, moved the logic to the `julia_flags()` function, updated the help string on `static_julia()`, and added a unit-test specifically covering `--cc-flags`. :)

PTAL. thanks again!

PR #46, opened 2018-04-03
Keep `builddir` relative to user's cwd.

NHDaly:
Before this change, a user-specified builddir would be evaluated as relative
to the juliaprog file. After this change, the builddir would be relative
to user's terminal's pwd.

Example:

```
$ cd ~/myApp
$ julia ~/.julia/v0.6/PackageCompiler/juliac.jl -vaej test/mytest.jl ~/.julia/v0.6/PackageCompiler/examples/program.c build
```

Before:
build directory = `~/myApp/test/build/mytest`
After:
build directory = `~/myApp/build/mytest`

--------------

If builddir is not specified, this commit will keep the default builddir the
same: `"$(dirname($juliaprog))/$builddir"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,9 +76,7 @@ positional arguments:
   cprog                 C program to compile (required only when
                         building an executable; if not provided a
                         minimal driver program is used)
-  builddir              directory used for building, either absolute
-                        or relative to the Julia program directory
-                        (default: "builddir")
+  builddir              build directory (default: "builddir")
 
 optional arguments:
   -v, --verbose         increase verbosity
--- a/src/static_julia.jl
+++ b/src/static_julia.jl
@@ -59,7 +59,8 @@ compiles the Julia file at path `juliaprog` with keyword arguments:
 """
 function static_julia(
         juliaprog;
-        cprog = nothing, builddir = "builddir", juliaprog_basename = splitext(basename(juliaprog))[1],
+        cprog = nothing, juliaprog_basename = splitext(basename(juliaprog))[1],
+        builddir = "$juliaprog_basename/builddir",
         verbose = false, quiet = false, clean = false,
         autodeps = false, object = false, shared = false, executable = false, julialibs = false,
     	sysimage = nothing, compile = nothing, cpu_target = nothing,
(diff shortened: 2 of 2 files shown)


SimonDanisch:
I guess the default behaviour is debatable.
But I guess it makes sense, especially if you want to have a `src` dir and a `build` dir in myapp...

NHDaly:
Yeah I wasn't sure about the default, either.

I can see it making sense if you're compiling from a totally different directory, `juliac.jl -vaej ~/Documents/code/myApp/mycode.jl`, then it makes sense for the default builddir to be in `myApp`.

lucatrv:
Yes, I gave `juliac.jl` that default behavior because I though the most common situation would be to run it from a different working directory and to customize either the build directory name and/or location relative to the Julia program. This is described by the `-h` documentation string "either absolute or relative to the Julia program directory" (but maybe it could be better explained). Of course we can decide to change that default behavior if useful / preferable.

NHDaly:
Oh, gotcha. Thanks for the explanation, @lucatrv.

My vote would be to switch the behavior to how I've implemented it in this PR:
 1. without specifying a builddir, it defaults to a directory called `builddir` next to the `julia_main`.
 2. if a relative-path builddir is specified, it is created relative to the current working directory (`pwd`) of the shell.

I feel strongly about point 2., because I think that is the most idiomatic with other unix tools. As one example, this is how gcc works: `gcc src/test.c -o build/out` will put `out` in `./build`, not `./src/build`.

I feel less strongly about what the best behavior is for point 1., and so I will defer to your judgement that defaulting to put it next to the julia source program is the best choice. I think that probably makes sense. But it's worth pointing out that here too, `gcc` defaults to putting it in your current directory (`gcc src/test.c; ./a.out`), so I could see that being preferable as well. what do you all think?

---------

Once we decide, I'll update the docs before this should be merged.

lucatrv:
Ok, you convinced me about point 2, but in this case I think also point 1 should be changed (default is `builddir` relative to the current working directory).

Note: my model for `juliac.jl` was actually [`meson`](http://mesonbuild.com/) rather than `gcc`.

NHDaly:
Oh huh, interesting. I've actually never heard of meson, but it looks neat.
Thanks for the link.

Okay, so basically we should change point 1 to be consistent, you think? So
everything will always be relative to the cwd, following the behavior of
`gcc`.

lucatrv:
Yes, I think we should change also point 1, to make everything relative to the cwd.

Note: Meson is now used by many [projects](http://mesonbuild.com/Users.html), and growing.

NHDaly:
👍 Thanks for all those changes as well, that looks good. :)

Issue #47, opened 2018-04-03
`juliac` fails to compile Blink.jl: `JULIA_HOME not defined`


NHDaly:
`juliac` is failing when I try to compile a file that uses `Blink.jl`, complaining that `LoadError: UndefVarError: JULIA_HOME not defined`.

Here is the very simple file I was compiling:
```
# test/blink_test.jl

using Blink

Base.@ccallable function julia_main(args::Vector{String})::Cint
    win = Window(shell(), Dict(:width=>1100))
    return 0
end
```

And here's the failure:

```
  (11:17:03) $ julia ~/.julia/v0.6/PackageCompiler/juliac.jl -vaej test/blink_test.jl ~/.julia/v0.6/PackageCompiler/examples/program.c builddir
Julia program file:
  "/Users/daly/Documents/developer/code/deep/JuliaDeep/test/blink_test.jl"
C program file:
  "/Users/daly/.julia/v0.6/PackageCompiler/examples/program.c"
Build directory:
  "/Users/daly/Documents/developer/code/deep/JuliaDeep/test/builddir"
Make build directory
Change to build directory
Build module image files ".ji" in directory "/Users/daly/Documents/developer/code/deep/JuliaDeep/test/builddir":
  `/Applications/Julia-0.6.app/Contents/Resources/julia/bin/julia -Cnative -J/Applications/Julia-0.6.app/Contents/Resources/julia/lib/julia/sys.dylib --compile=yes --depwarn=yes --startup-file=no --precompiled=no --compilecache=no -e 'empty!(Base.LOAD_CACHE_PATH) # reset / remove any builtin paths
        push!(Base.LOAD_CACHE_PATH, abspath("/Users/daly/Documents/developer/code/deep/JuliaDeep/test/builddir")) # enable usage of precompiled files
        include("/Users/daly/Documents/developer/code/deep/JuliaDeep/test/blink_test.jl") # include Julia program file
        empty!(Base.LOAD_CACHE_PATH) # reset / remove build-system-relative paths'`
Build object file "blink_test.o" in directory "/Users/daly/Documents/developer/code/deep/JuliaDeep/test/builddir":
[...]

PR #50, opened 2018-04-08
Reenable `--precompiled=yes` module optimization.

NHDaly:
`--compilecache` and `--precompiled` were originally disabled to fix a
segfault, but only disabling `--compilecache` is sufficient to fix that
problem. This PR reenables `--precompiled=yes` as long as the build
is not using a custom `cpu_target`.

See this discussion for more details:
https://github.com/JuliaLang/PackageCompiler.jl/issues/47#issuecomment-379528765


NHDaly:
/cc: @lucatrv When you get a chance, it looks like this can be reenabled and will be a performance boost. :)

lucatrv:
Please find my comment [here](https://github.com/JuliaLang/PackageCompiler.jl/issues/47#issuecomment-380243538).

PR #57, opened 2018-05-05
Set PROGRAM_FILE global via `argv[0]`.

### LINKED ISSUE #54: `PROGRAM_FILE` not set when compiled through `juliac.jl`
`Base.PROGRAM_FILE` is empty when compiled through `juliac.jl`. (The global `ARGS` is empty, which is where `PROGRAM_FILE` is set from.)

`PROGRAM_FILE` is set [here](https://github.com/JuliaLang/julia/blob/v0.6.2/base/client.jl#L301), by popping the first value off of `ARGS`, iff `ARGS` is not empty (checked [here](https://github.com/JuliaLang/julia/blob/33fdd3/base/client.jl#L271)).

Currently, the default `program.c` skips setting the global `ARGS`, opting instead to [pass an Array to `julia_main`](https://github.com/JuliaLang/PackageCompiler.jl/blob/b8f0d0a/examples/program.c#L32-L40). This means that there is no way to access `PROGRAM_FILE` from compiled code.

NHDaly:
Changes default `program.c` to manually set `PROGRAM_FILE` via its `argv[0]`.

Ideally, I think we would also set global `ARGS`, in the same way the julia interpreter does, but I still haven't figured out how to get that to work.

For #54.

### DIFF
--- a/examples/program.c
+++ b/examples/program.c
@@ -21,24 +21,32 @@ int main(int argc, char *argv[])
     int retcode;
     int i;
     uv_setup_args(argc, argv); // no-op on Windows
+
     // initialization
     libsupport_init();
+
     // jl_options.compile_enabled = JL_OPTIONS_COMPILE_OFF;
     // JULIAC_PROGRAM_LIBNAME defined on command-line for compilation
     jl_options.image_file = JULIAC_PROGRAM_LIBNAME;
     julia_init(JL_IMAGE_JULIA_HOME);
 
-    // build arguments array: `String[ unsafe_string(argv[i]) for i in 1:argc ]`
-    jl_array_t *ARGS = jl_alloc_array_1d(jl_apply_array_type(jl_string_type, 1), 0);
-    JL_GC_PUSH1(&ARGS);
+    // Initialize Core.ARGS with the full argv.
+    jl_set_ARGS(argc, argv);
+
+    // Set PROGRAM_FILE to argv[0].
+    jl_set_global(jl_base_module,
+        jl_symbol("PROGRAM_FILE"), (jl_value_t*)jl_cstr_to_string(argv[0]));
+
+    // Set Base.ARGS to `String[ unsafe_string(argv[i]) for i in 1:argc ]`
+    jl_array_t *ARGS = (jl_array_t*)jl_get_global(jl_base_module, jl_symbol("ARGS"));
     jl_array_grow_end(ARGS, argc - 1);
     for (i = 1; i < argc; i++) {
         jl_value_t *s = (jl_value_t*)jl_cstr_to_string(argv[i]);
         jl_arrayset(ARGS, s, i - 1);
     }
+
     // call the work function, and get back a value
     retcode = julia_main(ARGS);
-    JL_GC_POP();
 
     // Cleanup and gracefully exit
     jl_atexit_hook(retcode);
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -49,6 +49,45 @@ end
     end
 end
 
+@testset "program.c" begin
+    @testset "args" begin
+        basedir = mktempdir();
+        argsjlfile = mktemp(basedir)[1];
+        write(argsjlfile, raw"""
+            Base.@ccallable function julia_main(argv::Vector{String})::Cint
+                println("@__FILE__: $(@__FILE__)")
+                println("PROGRAM_FILE: $(PROGRAM_FILE)")
+                println("argv: $(argv)")
+                # Sometimes code accesses ARGS directly, as a global
+                println("ARGS: $ARGS")
+                println("Base.ARGS: $(Base.ARGS)")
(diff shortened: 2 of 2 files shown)


NHDaly:
It's worth mentioning that in normal julia programs, `PROGRAM_FILE` is set by popping the first arg of the global `ARGS`:
https://github.com/JuliaLang/julia/blob/33fdd3/base/client.jl#L289

I tried just calling [`jl_set_ARGS(argc, argv)`](https://github.com/JuliaLang/julia/blob/v0.6.2/src/jloptions.c#L516), but no matter where I put it in `program.c`, the global `ARGS` was never set. Perhaps this is a bug in julia?

lucatrv:
Before merging this PR, maybe you should verify if it is actually due to a Julia bug, either on Discourse, or otherwise directly opening an issue on Julia's repository if you believe it is a bug. What do you think?

ihnorton:
Note that the assignment block in `jl_set_ARGS` is only executed if `jl_core_module` is initialized (`jl_core_module != NULL`). I don't know much about the structure of this package, but from a quick glance at program.c, I don't see any explicit initialization of that variable.

NHDaly:
Hmm, yeah i saw that, but it seems that yes, `jl_core_module` is initialized. I walked through with `lldb`, and it doesn't take the branch, since `jl_core_module` has a value, and so it does in fact execute the assignment block. Sorry, i should've mentioned that earlier.
In fact, `args` is not `null`, and it calls `jl_arrayset` for each argument supplied. Which is why it baffled me that there still isn't anything in the julia `ARGS` afterwards.

For completeness, i think the core_module is initialized here:
https://github.com/JuliaLang/julia/blob/master/src/staticdata.c#L1511

NHDaly:
I'll ask around on slack and/or file a bug. thanks for the help!

NHDaly:
Okay, so i've found the problem! It was not a bug, it turns out, but rather it's built-in behavior that the REPL does, which we weren't replicating.

Basically, we were reading from `Base.ARGS` (global `ARGS` points there). But `jl_set_ARGS` only sets the _Core.ARGS_!

In a normal program, the REPL automatically copies `Core.ARGS` into `Base.ARGS`, but we weren't getting that behavior.

ARGS are set here:
https://github.com/JuliaLang/julia/blob/v0.6.3/base/client.jl#L367-L368
and the PROGRAM_FILE is set here:
https://github.com/JuliaLang/julia/blob/v0.6.3/base/client.jl#L301

But `_start()` is only called from the repl, here:
https://github.com/JuliaLang/julia/blob/v0.6.3/ui/repl.c#L121


--------------

So we just need to either replicate that behavior from `program.c`, or petition Julia to set both Base.ARGS and Core.ARGS in `jl_set_ARGS`, or both.

examples/program.c
```diff
@@ -36,6 +44,37 @@ int main(int argc, char *argv[])
...
+    // Set Base.ARGS to argv[1:end], to match behavior of Base._start().
+    //  -- jl_set_const doesn't work because of https://github.com/JuliaLang/julia/pull/27520.
+    //jl_set_const(jl_base_module, jl_symbol("ARGS"), (jl_value_t*)ARGS);
+    //  -- This should work, but for some reason it doesn't. It says it's
+    //  -- "WARNING: redefining constant ARGS", but for w/e reason doesn't...
+    //  -- I don't understand why it's not working.
+    //jl_checked_assignment(jl_get_binding_wr(jl_base_module, jl_symbol("ARGS")),
+    //                       (jl_value_t*)ARGS);
+    // -- Trying this instead, even though it's bulky:
+    set_julia_base_ARGS(argc, argv);
```
NHDaly: So all of the above is really wonky... I first just tried overwriting the Base.ARGS via `jl_set_const`, but ran into this problem:
https://github.com/JuliaLang/julia/pull/27520

So then I tried overwriting it via `jl_checked_assignment`, which really should work, but for some reason the compiled program still has an empty `String[]` for ARGS, and i don't know why...

So finally, i've caved and i'm just manually appending all the fields to the Base.ARGS, just like `Base._start()` does, via this helper function `set_julia_base_ARGS`.

NHDaly:
(And we should squash at least the first two commits, before merging; i only left the first one there for review clarity. 😊)

lucatrv:
CI tests failing on Windows...

NHDaly:
Oops, thanks, I forgot about the `executable_ext` for windows. Repushing now...

Also, I took this chance to squash & reword the commits. :)

lucatrv:
Still failing...

NHDaly:
gah, sorry. i don't have access to a windows machine so it's hard for me to test... i'll see what i can do...

NHDaly:
Okay finally fixed!! I left the println in the test because actually I think it's useful overall.

Thanks for reminding me when it was failing! :)

lucatrv:
Merged!

NHDaly:
:) thanks!

PR #61, opened 2018-05-06
Add `--builddir` and `--outname` flags, remove `builddir` argument

lucatrv:


lucatrv:
By replacing the previous `builddir` positional argument with the new `--dir` flag, it is now possible to pass a custom build directory without necessarily passing also a custom `cprog` argument.

@NHDaly if you agree on this change I would consider it also for [ApplicationBuilder.jl](https://github.com/NHDaly/ApplicationBuilder.jl).

lucatrv:
Yes I am also not really happy with the flag names. I thought a lot about either `--builddir` / `--basename` or `--dir` / `--name`. I chose the latter ones only because they begin with `-d` / `-n` which I have to use as short flags to avoid overlaps (unless you have better suggestions).
Now that you say it, I am thinking that maybe `--builddir` / `--basename` or `--builddir` / `--buildname` would have actually been a better choice (still keeping `-d` / `-n` as short flags), what do you think?

NHDaly:
Yeah that makes sense.

Hmmm I think there are several good choices. `--builddir` and `--buildname`
seems reasonable.

I think actually having had more time to think about it, my current
favorite are `--outdir` and `--outname`, but I don't feel very strongly one
way or the other.

And yeah I think -d and -n make sense considering the options. Another very
reasonable option would be simply to not have a shortname. I don't think
either one is so obvious as to be necessary. Also, I imagine both will be
somewhat uncommon. What do you think about leaving off the shortnames?

lucatrv:
Ok, for now I updated the PR with `--builddir` / `--outname`, but I am going to think a little more about it. However I would not like to change the `--builddir` flag, because it would still default to `builddir` (same build directory name used by [Meson](http://mesonbuild.com/)).

NHDaly:
I think those sound great! :)

lucatrv:
So I merged it, thanks for the advice! :)

PR #62, opened 2018-05-10
Add other Julia compilation flags

lucatrv:


lucatrv:
@SimonDanisch @NHDaly 
this is a quite extensive PR, please review it, if you agree I will finalize and merge it.

NHDaly:
I'm interested in the use case for disabling/enabling `--precompiled` and `--compilecache`. Is there a reason that a user of this package would want that? If so, could we maybe also explain that a bit in the docstring?

juliac.jl
```diff
@@ -50,45 +50,69 @@ Base.@ccallable function julia_main(args::Vector{String})::Cint
...
+            help = "enable/disable incremental precompilation of modules"
+        "--home", "-H"
+            arg_type = String
+            metavar = "<dir>"
+            help = "set location of `julia` executable"
+        "--startup-file"
+            arg_type = String
+            metavar = "{yes|no}"
+            range_tester = (x -> x ∈ ("yes", "no"))
+            help = "load ~/.juliarc.jl"
```
NHDaly: From the arg name, `"--startup-file"`, I would assume that this flag allows me to _set a custom_ startup file. Is that desirable? I would either consider changing the behavior to what I just described, or changing the name of the variable to something like `--use-startupfile`?

NHDaly: From the arg name, `"--startup-file"`, I would assume that this flag allows me to _set a custom_ startup file. Is that desirable? I would either consider changing the behavior to what I just described, or changing the name of the variable to something like `--use-startupfile`?

src/static_julia.jl
```diff
@@ -156,36 +162,43 @@ function julia_flags(optimize, debug, cc_flags)
...
 
 function build_julia_cmd(
-        sysimage, compile, cpu_target, optimize, debug, inline, check_bounds,
-        math_mode, depwarn, startupfile = false
+        sysimage, precompiled, compilecache, home, startup_file, handle_signals,
+        compile, cpu_target, optimize, debug, inline, check_bounds, math_mode, depwarn
     )
+    # TODO: these may just be temporary workarounds, see: https://github.com/JuliaLang/PackageCompiler.jl/issues/47
+    precompiled == nothing && cpu_target == nothing || (precompiled = "no")
+    compilecache == nothing && (compilecache = "no")
```
NHDaly: hmm, i could be misunderstanding, but the first one doesn't feel correct to me. 
If the user passes `precompiled = "yes"`, won't this set precompiled to "no"?

I think maybe you want parentheses around the second two phrases?
```
 precompiled == nothing && (cpu_target == nothing || (precompiled = "no"))
```
But honestly that's still hard to parse. perhaps this can simply be an `if`-statement? I think this is what you meant?:
```
if (precompiled == nothing && cpu_target != nothing) precompiled = "no" end
# or this:
(precompiled == nothing && cpu_target != nothing) && (precompiled = "no")
```

NHDaly: hmm, i could be misunderstanding, but the first one doesn't feel correct to me. 
If the user passes `precompiled = "yes"`, won't this set precompiled to "no"?

I think maybe you want parentheses around the second two phrases?
```
 precompiled == nothing && (cpu_target == nothing || (precompiled = "no"))
```
But honestly that's still hard to parse. perhaps this can simply be an `if`-statement? I think this is what you meant?:
```
if (precompiled == nothing && cpu_target != nothing) precompiled = "no" end
# or this:
(precompiled == nothing && cpu_target != nothing) && (precompiled = "no")
```

lucatrv:
> I'm interested in the use case for disabling/enabling --precompiled and --compilecache. Is there a reason that a user of this package would want that?

We are still in an early stage of development of Static Julia, there are still doubts for instance whether disabling `compilecache` is required just because of a `julia` bug or other. Moreover things may change once `julia-compile` is [introduced](https://github.com/JuliaLang/julia/issues/15864), which may provide behaviors specifically addressed at static compilation.
So my intent is for now to provide full testing capabilities, and to keep `juliac` aligned as much as possible with `julia` (now) and with `julia-compile` (in future), so that it is possible to verify all possible flags combinations, while keeping safe defaults.

> If so, could we maybe also explain that a bit in the docstring?

I agree that in future, once everything is clarified, we could add explanations in the docstring, but for now I thought to leave `juliac` docstring as aligned as possible with `julia`, so that we can easily follow its development changes, until everything is clarified.

juliac.jl
lucatrv: As explained below, `juliac` provides same compilation flags as `julia` (in future `julia-compile`). See `julia -h`.

src/static_julia.jl
```diff
@@ -156,36 +162,43 @@ function julia_flags(optimize, debug, cc_flags)
 end
 
 function build_julia_cmd(
-        sysimage, compile, cpu_target, optimize, debug, inline, check_bounds,
-        math_mode, depwarn, startupfile = false
+        sysimage, precompiled, compilecache, home, startup_file, handle_signals,
+        compile, cpu_target, optimize, debug, inline, check_bounds, math_mode, depwarn
     )
+    # TODO: these may just be temporary workarounds, see: https://github.com/JuliaLang/PackageCompiler.jl/issues/47
```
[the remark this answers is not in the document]
lucatrv: Ok I'll change that.

src/static_julia.jl
lucatrv: In Julia (like in Bash) `&&` is passed through only if the previous expression returns `true`, while `||` is passed through only if the previous expression returns `false`. So if `precompiled = "yes"`, then `precompiled == nothing` returns `false`, and therefore `&&` is not passed through.

juliac.jl
NHDaly: okay i'm convinced. :) thanks!

NHDaly: (sorry i misunderstood that what you were doing in this PR is just copying forward all of the julia flags! Makes sense.)

src/static_julia.jl
NHDaly: :) So I think everything you just said is correct, and yet, behold:
```
julia> false && true || println("HI")
HI
```
I think the nuance is here:
> while `||` is passed through only if the previous expression returns `false`.

In this case, the _previous expression_ is actually _the truth-value on the left-hand side of the ||_. I would imagine it like this:
```
`precompiled == nothing` --> <evaluate expression ... okay, it's false!> --> `false` --> `&& (cpu_target == nothing)` --> <okay, skip the right-hand side, cause this already evaluates to false --> `false` --> `|| (precompiled = "no")` --> <now we need to evaluate the right-hand side again because the left-hand side was false, so the truthiness of the whole expression depends on the right-hand side> --> (precompiled = "no") --> <returns `nothing`>.
``` 

That's why I suggested, if you want to keep it exactly as you currently have it, you need to put parens around the entire right-hand side _of `&&`_, so that it can skip the whole thing:
```
julia> false && (true || println("hello?"))
>> false
```
But i still think an `if`-statement might be better, since, clearly, multiple short-circuiting statements can be confusing and hard to read. :)

NHDaly: (In bash as well):
```
$ false && true || echo "HI"
HI
$ false && (true || echo "helloooooooooo")
$
```

lucatrv: You're perfectly right of course :) I'll fix that, thanks!

lucatrv: This is pretty straightforward:
```
precompiled == nothing && cpu_target != nothing && (precompiled = "no")
```

NHDaly:
👍👍👍

PR #70, opened 2018-05-19
Do not change directory when building

lucatrv:


lucatrv:
@NHDaly this PR passes all tests on Linux and Windows, but not on macOS. I cannot figure out why, also because I do not have any macOS system available, can you please verify it on your system? Thanks

NHDaly:
:) Yeah I'll take a look tomorrow morning! :)

src/static_julia.jl
```diff
@@ -209,20 +202,20 @@ function build_object(
         sysimage, precompiled, compilecache, home, startup_file, handle_signals,
         compile, cpu_target, optimize, debug, inline, check_bounds, math_mode, depwarn
     )
+    cache_dir = joinpath(dirname(o_file), "cache_ji_v$VERSION")
```
NHDaly: Any reason you dropped the `abspath` here? Maybe it's not needed?

lucatrv:
Unfortunately `abspath` did not fix the issue...

NHDaly:
Okay, so, here's the problem:

```
shell> ./hello
dyld: Library not loaded: @rpath//var/folders/5j/c6kd481j4l71m802wh5qp6vc0000gn/T/tmp8Zer00/hello.dylib
  Referenced from: /private/var/folders/5j/c6kd481j4l71m802wh5qp6vc0000gn/T/tmp8Zer00/./hello
  Reason: image not found
```

That is, two problems:
 1. for some reason the .dylib's absolute path is being considered as a relative path when linking the binary. It's looking for the binary here:
    ```
    $ otool -L hello
    hello:
        @rpath//var/folders/5j/c6kd481j4l71m802wh5qp6vc0000gn/T/tmpXUWsRI/hello.dylib (compatibility version 0.0.0, current version 0.0.0)
    ```
2. the .dylib also believes it will be located at that "relative" path:
    ```
    $ otool -L hello.dylib
    hello.dylib:
        @rpath//var/folders/5j/c6kd481j4l71m802wh5qp6vc0000gn/T/tmpXUWsRI/hello.dylib (compatibility version 0.0.0, current version 0.0.0)
    ```

To fix it, you'll have to use `install_name_tool` to fix the .dylib's believed destination and to fix where the binary is looking. You need both of these commands for it to work:
```
$ install_name_tool -id '@rpath/hello.dylib' hello.dylib
$ install_name_tool -change '@rpath//var/folders/5j/c6kd481j4l71m802wh5qp6vc0000gn/T/tmpXUWsRI/hello.dylib' '@rpath/hello.dylib' hello
```

I haven't yet been able to find a way to fix this just through `gcc` itself other than `cd`ing into the destination when building (which is what you were trying to avoid).

NHDaly:
Also, I've opened https://github.com/lucatrv/PackageCompiler.jl/pull/1 to switch the order that the `runtests` testsets run in which should make it easier for you to see what's broken on Travis.

NHDaly:
Ah, okay, it was actually a very simple fix:
https://github.com/lucatrv/PackageCompiler.jl/pull/2/files#diff-d47d0c2f9ab34ed28f73bdd671edd18cR235

You just missed a `basename` in the `if is_apple()` section of building the shared library. :)

I've opened https://github.com/lucatrv/PackageCompiler.jl/pull/2 to send up the fix. :)

lucatrv:
Great!!

NHDaly:
Is this ready to submit? It looks good to me!

lucatrv:
Yes sorry, I have just added a last commit, I am going to merge it if all tests pass.

PR #72, opened 2018-05-27
Fix path of precompiled.jl and library as included from julia_main.jl

tkoolen:
Since `julia_main.jl` lives in `builddir`, it should include files relative to `builddir`.

Without this change, something similar to `include("builddir/precompiled.jl")` would appear in `builddir/julia_main.jl`, resulting in `ERROR: LoadError: could not open file`.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -24,21 +24,27 @@ julia = Base.julia_cmd().exec[1]
 end
 
 @testset "build_executable" begin
-    builddir = mktempdir()
     jlfile = joinpath(@__DIR__, "..", "examples", "hello.jl")
-    snoopfile = open(joinpath(builddir, "snoop.jl"), "w") do io
-        write(io, open(read, jlfile))
-        println(io)
-        println(io, "using .Hello; Hello.julia_main(String[])")
+    basedir = mktempdir()
+    relativebuilddir = "build"
+    cd(basedir) do
+        mkdir(relativebuilddir)
+        snoopfile = "snoop.jl"
+        open(snoopfile, "w") do io
+            write(io, open(read, jlfile))
+            println(io)
+            println(io, "using .Hello; Hello.julia_main(String[])")
+        end
+        build_executable(
+            jlfile, snoopfile = snoopfile, builddir = relativebuilddir, verbose = true
+        )
     end
-    build_executable(
-        jlfile, snoopfile = snoopfile, builddir = builddir, verbose = true
-    )
+    builddir = joinpath(basedir, relativebuilddir)
     @test isfile(joinpath(builddir, "hello.$(Libdl.dlext)"))
     @test isfile(joinpath(builddir, "hello$executable_ext"))
     @test success(`$(joinpath(builddir, "hello$executable_ext"))`)
     for i = 1:100
-        try rm(builddir, recursive = true) end
+        try rm(basedir, recursive = true) end
         sleep(1/100)
     end
 end
--- a/src/api.jl
+++ b/src/api.jl
@@ -79,8 +79,8 @@ function build_executable(
         snoop(snoopfile, precompfile, joinpath(builddir, "snoop.csv"))
         jlmain = joinpath(builddir, "julia_main.jl")
         open(jlmain, "w") do io
-            println(io, "include(\"$(escape_string(precompfile))\")")
-            println(io, "include(\"$(escape_string(library))\")")
+            println(io, "include(\"$(escape_string(relpath(precompfile, builddir)))\")")
+            println(io, "include(\"$(escape_string(relpath(library, builddir)))\")")
         end
         library = jlmain
     end


NHDaly:
Ah, so I think this is actually resulting from a confusion regarding the names of the options. `builddir` is actually the *output dir*, that is, it's where all the *built artifacts* are placed. So you shouldn't have to change any `include` statements in your code at all in order to compile it; everything should work the same whether it's invoked through `julia`, or compiled through `juilac.jl`.

You can assume all the include-paths will be relative to the `julia_main.jl` file you pass to `juliac`.

I think the most common pattern is to use a separate builddir for the build artifiacts, and to  *not* have any code inside the builddir.

For example, if my code is here:
```
MyPackage/
  ↳ julia_main.jl   # contains `include("utils.jl")`
  ↳ utils.jl
```
Then I can build that code with the following command:
```
$ cd MyPackage
$ julia ~/.julia/v0.6/PackageCompiler/juliac.jl -vae julia_main.jl --builddir outbuild --outname mypackage
```
Which creates the `outbuild` directory and puts all the built objects inside it:
```
MyPackage/
  ↳ outbuild/
      ↳ mypackage
      ↳ mypackage.dylib
  ↳ julia_main.jl
  ↳ utils.jl
```

-------------------

If instead you were trying to point `juliac` at code that isn't located at your current directory, you can just pass the full or relative path to your code, like so:
```
$ cd ~
$ julia ~/.julia/v0.6/PackageCompiler/juliac.jl -vae ~/Documents/developer/code/games/pongClone/main.jl --builddir ./outbuild
$ ls ~/outbuild
main  main.dylib main.o
```
Does that make sense?

So it's *possible* to use the same directory for the builddir and the code, but that's not as common since it is harder to clean up.

tkoolen:
Note that this is a change to `build_executable`; I'm not running `juliac.jl` (just following the usage example in the readme). If I shouldn't be calling `build_executable`, then I agree that this is a documentation issue.

If `snoopfile` is not `nothing`, `build_executable` creates its own `julia_main.jl` file inside the build directory, so I don't have control over where that's being put.

https://github.com/JuliaLang/PackageCompiler.jl/blob/b69425a3892e8f9e7707823cc0327d4936b68c7a/src/api.jl#L77-L86

Here's my current setup: https://github.com/JuliaRobotics/RigidBodyDynamics.jl/tree/tk/static/static. Note that my `builddir` (called `build`) will be created as a subdirectory of the `static` directory and passed into `build_executable` as a path relative to the `static` directory (i.e., not the _same_ as the `static` directory that contains the 'main' to compile).

NHDaly:
Oh I see. I'm sorry, i completely misunderstood the context for this change. Please ignore my previous message.

Yes, I think you are right. This seems like a necessary fix.


Also, we should have unit tests for this case. I personally haven't ever used a snoopfile with `build_executable` so i'm not really familiar with how it works. Would you be able to add a test as well?

tkoolen:
While trying to cause a test failure with the old code I found out that there's a bug in the `build_executable` test set, namely that

https://github.com/JuliaLang/PackageCompiler.jl/blob/4d05f6bef013a0bec7eadc1ab79d0625bb1f887a/test/runtests.jl#L29-L33

returns `nothing` (instead of the file name). As a result, the test didn't actually cover the snoopfile code in `api.jl`  (witnessed by https://codecov.io/gh/JuliaLang/PackageCompiler.jl/src/master/src/api.jl).

Fixing it now.

tkoolen:
OK, updated the `build_executable` test set to fix the test bug and so that it triggers the bug that this PR is intended to solve on current master.

test/runtests.jl
```diff
@@ -24,21 +24,27 @@ julia = Base.julia_cmd().exec[1]
...
+        snoopfile = "snoop.jl"
+        open(snoopfile, "w") do io
+            write(io, open(read, jlfile))
+            println(io)
+            println(io, "using .Hello; Hello.julia_main(String[])")
+        end
+        build_executable(
+            jlfile, snoopfile = snoopfile, builddir = relativebuilddir, verbose = true
+        )
     end
```
NHDaly: I think @lucatrv is trying to remove all `cd` from the package. Can you undo that part of the change? You'll just have to have more `joinpath`s. Thanks! :)

test/runtests.jl
```diff
@@ -24,21 +24,27 @@ julia = Base.julia_cmd().exec[1]
...
 
 @testset "build_executable" begin
-    builddir = mktempdir()
     jlfile = joinpath(@__DIR__, "..", "examples", "hello.jl")
-    snoopfile = open(joinpath(builddir, "snoop.jl"), "w") do io
-        write(io, open(read, jlfile))
-        println(io)
-        println(io, "using .Hello; Hello.julia_main(String[])")
+    basedir = mktempdir()
+    relativebuilddir = "build"
```
NHDaly: I think it makes sense to separate the `snoopfile` from the `builddir` as you've done here though. thanks.

test/runtests.jl
tkoolen: I think removing `cd`s from the production code is great, but the `cd` here is kind of required to accurately reproduce what at least in my opinion is a pretty standard setup (which triggers the original issue), while avoiding creating files inside PackageCompiler.jl's test directory.

lucatrv:
All this looks fine to me, but why is it now triggering a build error on win32?

tkoolen:
Seems to be an issue while installing `gcc`?

```
7-Zip 9.20  Copyright (c) 1999-2010 Igor Pavlov  2010-11-18
Processing archive: C:\Users\appveyor\.julia\v0.6\WinRPM\cache\1\noarch%2Fmingw32-zlib1-1.2.8-9.17.noarch.rpm
Error: Can not open file as archive
===========================[ ERROR: PackageCompiler ]===========================
LoadError: failed process: Process(`'C:\projects\julia\bin\7z.exe' x -y 'C:\Users\appveyor\.julia\v0.6\WinRPM\cache\1\noarch%2Fmingw32-zlib1-1.2.8-9.17.noarch.rpm' '-oC:\Users\appveyor\.julia\v0.6\WinRPM\cache\1'`, ProcessExited(2)) [2]
while loading C:\Users\appveyor\.julia\v0.6\PackageCompiler\deps\build.jl, in expression starting on line 63
================================================================================
```

The `noarch%2F` looks a little strange.

lucatrv:
Yeah I think it's just a false alarm, so I'm gonna merge this PR.

tkoolen:
OK, great. Thanks for the cool package!

PR #80, opened 2018-06-14
Have all static_julia apis return true on success.

NHDaly:
Change `build_sysimg`, `build_shared_lib`, and `build_executable` to
return `true` on success to allow `&&` chaining them with other commands
e.g., this used to fail:
julia> build_executable("test.jl")) && run("./builddir/test")
...
ERROR: TypeError: non-boolean (Void) used in boolean context


NHDaly:
So, actually, after thinking about this more, this maybe isn't a good idea, since the failure case doesn't return `false`, it throws an ErrorException...

So if you wrote

    julia> build_executable("test.jl")) || println("failure")

your print statement would never execute...
So maybe this isn't worth it..

PR #87, opened 2018-07-13
WIP: Julia v0.7 compatability for static_julia

NHDaly:
Trying to get julia v0.7 programs to compile as well! :)

There were a few small syntaxy changes, and now it's just some compilation functionality that's left.


NHDaly:
I *think* this is right. I'm pretty sure LOAD_CACHE_PATH was renamed to DEPOT_PATH, though the only documentation i can find on it is this:
Yeah okay i think i've answered my own question. If anyone else knows, though, please let me know! :blush:
```
julia v0.6>
help?> Base.compilecache
  Base.compilecache(module::String)

  Creates a precompiled cache file for a module and all of its dependencies. This can be used to reduce
  package load times. Cache files are stored in LOAD_CACHE_PATH[1], which defaults to
  ~/.julia/lib/VERSION. See Module initialization and precompilation for important notes.
```
```
julia v0.7>
help?> Base.compilecache
  Base.compilecache(module::PkgId)

  Creates a precompiled cache file for a module and all of its dependencies. This can be used to reduce
  package load times. Cache files are stored in DEPOT_PATH[1]/compiled. See Module initialization and
  precompilation for important notes.
```

NHDaly:
For now, I split off `program_v07.c`, but if there's a good way to check the julia version inside the C code that would work too. it might be `#DEFINED` somewhere in the julia headers somewhere -- i didn't look too hard.

NHDaly:
There's some progress here. I am able to compile little, useless files like this one now:
```julia
Base.@ccallable function julia_main()::Cint
    return 10
end
```
but that's not very interesting.

When I do more complicated things like `print`, it segfaults. 😬 I'm asking for help here:
https://github.com/JuliaLang/julia/issues/28117

------------

Then, also, a second problem is that when compiling, i can't get it to see any of the Packages I have installed. The spawned `julia` process gives up and complains that it can't find the Package. But I had it print out its `LOAD_PATH`, and that seemed fine, so I'm not sure what's wrong. Packages are loaded in a new way in `0.7` with `Pkg3`, and I'm not very familiar with it yet.

```julia
julia> mktemp() do f,_
           write(f, """using Glob
                       Base.@ccallable function julia_main()::Cint
                           return 0
                       end
                    """)
           build_executable(f, "test"; verbose=true);
           success(`builddir/test`)
       end
...
ERROR: LoadError: ArgumentError: Package Glob not found in current path:
 - Run `Pkg.add("Glob")` to install the Glob package.
...
```

NHDaly:
Closing this PR in favor of #88.

PR #88, opened 2018-07-19
Get `static_julia` working for simple files on julia v0.7!

NHDaly:
This PR improves `julia v0.7` compatibility for `static_julia`.

~It builds`examples/hello.jl` successfully, and a few other examples I've tried.~

EDIT: no it doesn't, I don't know why I thought that. 😊

### DIFF
--- a/src/static_julia.jl
+++ b/src/static_julia.jl
@@ -225,14 +225,20 @@ function build_object(
 end
 
 function build_shared(s_file, o_file, verbose, optimize, debug, cc, cc_flags)
+    # Prevent compiler from stripping all symbols from the shared lib.
+    if julia_v07
+        if isapple()
+            o_file = `-Wl,-all_load $o_file`
+        else
+            o_file = `-Wl,--whole-archive $o_file -Wl,--no-whole-archive`
+        end
+    end
     command = `$cc -shared -o $s_file $o_file $(julia_flags(optimize, debug, cc_flags))`
     if isapple()
         command = `$command -Wl,-install_name,@rpath/$(basename(s_file))`
     elseif iswindows()
         command = `$command -Wl,--export-all-symbols`
     end
-    # Prevent compiler from stripping all symbols from the shared lib.
-    julia_v07 && (command = `$command -Wl,-$(isapple() ? "all_load" : "whole-archive")`)
     verbose && println("Build shared library \"$s_file\":\n  $command")
     run(command)
 end
--- a/src/static_julia.jl
+++ b/src/static_julia.jl
@@ -203,17 +203,15 @@ function build_object(
         expr = "
   Base.init_depot_path() # initialize package depots
   Base.init_load_path() # initialize location of site-packages
-  empty!(Base.DEPOT_PATH) # reset / remove any builtin paths
-  push!(Base.DEPOT_PATH, \"$cache_dir\") # enable usage of precompiled files
-  include(\"$juliaprog\") # include Julia program file
-  empty!(Base.DEPOT_PATH) # reset / remove build-system-relative paths"
+  Sys.__init__();  # Needed to find built-in Modules.
+  include(\"$juliaprog\") # include Julia program file"
     else
         expr = "
   empty!(Base.LOAD_CACHE_PATH) # reset / remove any builtin paths
   push!(Base.LOAD_CACHE_PATH, \"$cache_dir\") # enable usage of precompiled files
   Sys.__init__(); Base.early_init(); # JULIA_HOME is not defined, initializing manually
-  include(\"$juliaprog\") # include Julia program file
-  empty!(Base.LOAD_CACHE_PATH) # reset / remove build-system-relative paths"
+  include(\"$juliaprog\") # include julia program file
+  empty!(base.LOAD_CACHE_PATH) # reset / remove build-system-relative paths"
     end
     if compilecache == "yes"
         command = `$julia_cmd -e $expr`
--- a/src/static_julia.jl
+++ b/src/static_julia.jl
@@ -204,14 +204,15 @@ function build_object(
   Base.init_depot_path() # initialize package depots
   Base.init_load_path() # initialize location of site-packages
   Sys.__init__();  # Needed to find built-in Modules.
+  Base.__init__();
   include(\"$juliaprog\") # include Julia program file"
     else
         expr = "
   empty!(Base.LOAD_CACHE_PATH) # reset / remove any builtin paths
   push!(Base.LOAD_CACHE_PATH, \"$cache_dir\") # enable usage of precompiled files
   Sys.__init__(); Base.early_init(); # JULIA_HOME is not defined, initializing manually
-  include(\"$juliaprog\") # include julia program file
-  empty!(base.LOAD_CACHE_PATH) # reset / remove build-system-relative paths"
+  include(\"$juliaprog\") # include Julia program file
+  empty!(Base.LOAD_CACHE_PATH) # reset / remove build-system-relative paths"
     end
     if compilecache == "yes"
         command = `$julia_cmd -e $expr`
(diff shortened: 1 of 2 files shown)


NHDaly:
Okay! @lucatrv I think this should be enough for v0.7 support for static_julia! :) There's still more changes needed for `juliac.jl`, and neither snoop files nor the rest of PackageCompiler work I think.

But still, it's cool! :) This should be ready to merge, as far as i can tell.

NHDaly:
Okay so actually, this only works for basic files that don't `import` or `using` any Packages... When compiling, it can't find any packages, even though they're installed:
```julia
julia> build_executable("/Users/daly/.julia/v0.6/PackageCompiler/examples/hello.jl")
Julia program file:
  "/Users/daly/.julia/v0.6/PackageCompiler/examples/hello.jl"
C program file:
  "/Users/daly/.julia/v0.6/PackageCompiler/examples/program.c"
Build directory:
  "/Users/daly/.julia/v0.6/PackageCompiler/builddir"
WARNING: julia --compilecache option is deprecated, use --compiled-modules instead.
ERROR: LoadError: ArgumentError: Package UnicodePlots not found in current path:
 - Run `Pkg.add("UnicodePlots")` to install the UnicodePlots package.
```

I'm sure it's something to do with the LOAD_CACHE_PATH change.

lucatrv:
I will make some further attempts on this, but I guess there may still be open issues in Julia 0.7 preventing successful static compilation... maybe we should wait for final release?

NHDaly:
Hmm it's possible, yeah, but I'd bet there's just something we need to change in the way we do it. I've asked for help here: https://github.com/JuliaLang/julia/issues/28117.
I think it would be good to get it working, now, while 0.7 is still in beta. Because if it _is_ something wrong with the release, they'd want to know before it's officially released so that it can be fixed before the release! :)

NHDaly:
Okay actually, i've pushed up adding `Base.__init__()` as well. I think it seems reasonable to expect Base to be initialized. I know we ran into all kinds of problems in v0.6 because it wasn't.

We can always try removing it in the future as an optimization, but I think for correctness it should probably be there too.

Okay, take a look when you can, because I think this is all set! :)

NHDaly:
Cool!

Huh, I'm not sure what the windows equivalent is. Maybe just try without
it? Perhaps it's not even needed!

When I didn't use it, the shared library it produced was empty (like 4K).
Maybe try removing it to see if it just works? :)

lucatrv:
This is actually much better than I thought! So now with Julia 0.7 we get a static library (`.a` archive file), and therefore we should be able to do two things:

1. Statically compile a self contained executable (I plan to make this the current `-e` flag behavior).
2. Compile a shared library (from the static library), and then create an executable dynamically linked to it (I plan to add a new `-E` flag for this).

Unfortunately neither one of the two works for me (I'm on Linux), and I get `ERROR: invalid redefinition of constant UInt32`. However I think we are very close...

On Linux I can create a `.so` shared library from a `.a` static library with `-Wl,--whole-archive staic_libray.a -Wl,--no-whole-archive`. Can you please verify if this works also on macOS? Basically you would use in `build_shared`:

```
command = `$cc -shared -o $s_file -Wl,--whole-archive $o_file -Wl,--no-whole-archive $(julia_flags(optimize, debug, cc_flags))`
```

(and you would not add `-Wl,-all_load`)

lucatrv:
Ok, I managed to make method 2 work. So if you confirm whether `-Wl,--whole-archive $o_file -Wl,--no-whole-archive` works on macOS I could for now merge this. With method 1 (statically linked executable) I still get `ERROR: invalid redefinition of constant UInt32` but I guess we'll be able to fix it later...

NHDaly:
> 1. Statically compile a self contained executable (I plan to make this the current `-e` flag behavior).
> 1. Compile a shared library (from the static library), and then create an executable dynamically linked to it (I plan to add a new `-E` flag for this).

Ah, so, actually, I thought the same thing, but Jameson says it's not really worth the trouble. It might be possible, but at least for now it's not well supported and lots of stuff probably won't work correctly. (For example, I was able to get it to compile statically eventually, but if I tried to `println`, it would segfault).

You can read my failed attempts to get that to work, and Jameson's comments, here:
https://github.com/JuliaLang/julia/issues/28117

In particular this comment:
https://github.com/JuliaLang/julia/issues/28117#issuecomment-406436359

---

Sad, I know. :/

But that doesn't mean we can't keep trying down this path in the future, if it's something that interests you! For now, though, I think it would be best to get the regular sysimg mode working in `v0.7` first, then we can try other things after that.

Does that sound okay?

> So if you confirm whether -Wl,--whole-archive $o_file -Wl,--no-whole-archive works on macOS I could for now merge this. 
Ah, it sounds like `all_load` is the mac equivalent of `whole-archive`. So perhaps we need an `isapple` switch there to switch between them? I'll push that up now for you to look at.
https://stackoverflow.com/a/16234257/751061

NHDaly:
Okay, give this a try if you would! It should work on windows now hopefully! :)

lucatrv:
I assume that `-Wl,--whole-archive $o_file -Wl,--no-whole-archive` does not work on macOS, can you confirm? (otherwise we could simply use it on all platforms).

lucatrv:
Please notice that `-Wl,-whole-archive` does not work on Linux. First of all two dash characters are needed: `-Wl,--whole-archive`. Then the static library should be wrapped between `-Wl,--whole-archive` and `-Wl,--no-whole-archive`. So this is the correct command on Linux (I have not verified it on Windows yet, but I guess it should work as well):

```
command = `$cc -shared -o $s_file -Wl,--whole-archive $o_file -Wl,--no-whole-archive $(julia_flags(optimize, debug, cc_flags))`
```

Can you please verify if it works on macOS? Thanks

NHDaly:
Ah, sorry I didn't specifically respond to your question. Yes, that flag does not work on mac. On macOS, the equivalent flag is `-Wl,-all_load`:

```
  `cc -shared -o /Users/daly/.julia/dev/PackageCompiler/builddir/test.dylib -Wl,--whole-archive /Users/daly/.julia/dev/PackageCompiler/builddir/test.a -Wl,--no-whole-archive -std=gnu99 -I/Applications/Julia-0.7.app/Contents/Resources/julia/include/julia -DJULIA_ENABLE_THREADING=1 -fPIC -L/Applications/Julia-0.7.app/Contents/Resources/julia/lib -Wl,-rpath,/Applications/Julia-0.7.app/Contents/Resources/julia/lib -Wl,-rpath,/Applications/Julia-0.7.app/Contents/Resources/julia/lib/julia -ljulia -m64 -Wl,-install_name,@rpath/test.dylib`
ld: unknown option: --whole-archive
```

I didn't realize that the second half is needed, sorry. I'll be sure to add that now!

NHDaly:
Okay, I've updated it again to be what we've discussed: `-Wl,-all_load $o_file` on mac and `-Wl,--whole-archive $o_file -Wl,--no-whole-archive on windows and linux. :)

Sorry for my confusion earlier and thanks for working through this with me!

NHDaly:
Ah, sorry about the `v0.6` incompatibilities on the test file. I hadn't realized I had run `using Compat` locally before running the tests.

But yeah, those can all also be fixed with just `using Compat` as well, which might be easier than keeping all `if julia_v07` checks around.

------------

Glad we got this to work, though! Thanks again for your help! It's exciting! :)

lucatrv:
Thanks for your hint, but since Static Julia requires lots of testing and troubleshooting I prefer to keep track of both code versions myself. Once Julia v0.7 is released I will grep "07" and remove all legacy code.

NHDaly:
That make sense!

(Not right away though, hopefully! There's a lot of packages that are still
slow moving, and it would be nice to maintain backwards compatibility as
long as possible!)

PR #95, opened 2018-07-25
Add cc,cc_flags to build_executable,build_shared_lib apis

NHDaly:


lucatrv:
I agree with the changes to `api.jl`, but I think you can remove the changes to `static_julia.jl` as in my opinion it is ok to define defaults there, as it is also for `cprog`, `builddir`, `outname` (unless there is a reason that I cannot see right now).

NHDaly:
Ah, yeah, sorry, I should've explained myself.

I think the alternative to what I did would be just to use the default value for `cc` (`system_compiler`) in `api.jl`. But then I realized that `system_compiler` is a global defined in `static_julia.jl`, and I wasn't sure if it's bad style to use that everywhere? I guess if we change our mind about what the default should be now we have to change it in three places instead of one?

But if you think it's okay, i'll just reuse `system_compiler` in all those places then! :)

lucatrv:
I agree we should define all flag defaults in a single place (`static_julia.jl`), so I am going to fix this also for the other flags.

PR #100, opened 2018-07-29
Add `--copy-file` flag

lucatrv:


NHDaly:
:) Hey, Luca!

NHDaly:
:) Hey Luca!

EDIT: -- GAH -- sorry, that shift-enter auto-submit got me again...

I saw this PR go by. It looks similar to the `-R` / `--resource` flag we have on https://github.com/NHDaly/ApplicationBuilder.jl. For that flag, i also needed to specify lists of things, and instead of delimiting a list somehow, I ended up just doing repetitions. So `-R img.png -R music/` would copy both the music directory and img.png.

I thought that seemed a bit easier for users. What do you think?

Here's how i implemented it, for reference:
https://github.com/NHDaly/ApplicationBuilder.jl/blob/e11dcb5792f3ee75613a173ddfdac53f16d1db83/build_app.jl#L23-L29
https://github.com/NHDaly/ApplicationBuilder.jl/blob/e11dcb5792f3ee75613a173ddfdac53f16d1db83/src/BuildApp.jl#L84-L114

I have also been thinking that it might be nice to have the option of specifying a destination path, but so far i haven't seen a real need for that.

NHDaly:
:) Cool!

> As for the flag name though I prefer not to use -R because I plan to add a new --Release flag for fully automated release build (equivalent to -caetjr).

Oh, yeah no sorry. I agree the current flag name is fine. I didn't mean to suggest you use the same name, just that the implementation will probably be similar. :)

PR #113, opened 2018-09-03
Move to Julia 1.0

lucatrv:

PR #116, opened 2018-09-12
Move to Julia 1.0

vtjnash:
Commits from https://github.com/JuliaLang/PackageCompiler.jl/pull/113, plus ~one~some, rebased—now to see what CI says


lucatrv:
Eureka!! Many thanks for your effort @vtjnash !!!

SimonDanisch:
Yaaay, thanks :)

lucatrv:
@vtjnash since you have just been working on `PackageCompiler` could you please double check the following code in `static_julia.jl`:

```julia
cache_dir = "cache_ji_v$VERSION"
# TODO: verify if this initialization is correct for Julia v1.0
expr = """
    Base.__init__(); Sys.__init__() # initialize `Base` and `Sys` modules
    pushfirst!(Base.DEPOT_PATH, $(repr(cache_dir))) # save precompiled modules locally
    include($(repr(juliaprog))) # include Julia program file
    """
# TODO: fix this for Julia v1.0, or how to precompile modules?
#if compiled_modules == "yes"
#    command = `$julia_cmd -e $expr`
#    verbose && println("Build \".ji\" local cache:\n  $command")
#    cd(builddir) do
#        run(command)
#    end
#end
command = `$julia_cmd --output-o $o_file -e $expr`
```

This basically comes from the original code that you added to [`static-julia`](https://github.com/JuliaComputing/static-julia/pull/40), which I adapted to make it work (partially) with Julia 1.0. However modules do not seem to be correctly precompiled and cached in local directory `cache_dir`. Moreover we found issues with `sysimage_native_code=yes` (previously `precompile=yes`) when a custom `cpu_target` is declared, and with `compiled_modules=yes` (previously `precompile=yes`), so currently they are both disabled by default, see discussion [here](https://github.com/JuliaLang/PackageCompiler.jl/issues/47) and the following code in `static_julia.jl`:

```julia
sysimage_native_code == nothing && cpu_target != nothing && (sysimage_native_code = "no")
compiled_modules == nothing && (compiled_modules = "no")
```

Can you please give me some hints on this?
Thanks

vtjnash:
Yes, currently you need to do a double-step (`./julia -J file.so --output-ji file.ji`; `./julia -J file.ji -C arch --output-o file.o.a`) to switch cpu target reliably. In the future, should probably add that as a flag (`./julia --output-C native`) so it can be done in one step.

The cache directory got renamed to `DEPOT_PATH`, so you can't have a local directory just for build artifacts anymore, but the step to set DEPOT_PATH in the above snippet should suffice.

Issue #119, opened 2018-09-18
fails Pkg.test with Julia v1.0 


geoff2273:
ERROR: LoadError: UndefVarError: Test not defined
Stacktrace:
 [1] include at .\boot.jl:317 [inlined]
 [2] include_relative(::Module, ::String) at .\loading.jl:1038
 [3] include(::Module, ::String) at .\sysimg.jl:29
 [4] include(::String) at .\client.jl:388
 [5] top-level scope at none:0
in expression starting at C:\Users\GEvans\.julia\packages\PackageCompiler\5u3Km\test\runtests.jl:2

Need to replace line 2: "using Base.Test" with "using Test"

Now just need to find out why the test are failing...

geoff2273:
Also need to add "using Pkg" to the top of snooping.jl due to line 99: Pkg.dir(package)

This then raises a warning that Pkg isn't in the deps.

Then I get errors about joinpath: ERROR: MethodError: no method matching joinpath(::Nothing, ::String)

I think this package is not yet Julia v1.0 ready! I will check again in a few months.

lucatrv:
Julia 1.0 support is still WIP, see PR #113 and #116

lucatrv:
Julia v1.0 support added with PR #116

Issue #115, opened 2018-09-05
Can't install package


ViralBShah:
I am on PackageCompiler v0.4.0.

````
viralbshah@arctic1:~/julia-1.0.0$ ./bin/julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.0.0 (2018-08-08)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(v1.0) pkg> add PackageCompiler
   Cloning default registries into /home/viralbshah/.julia/registries
   Cloning registry General from "https://github.com/JuliaRegistries/General.git"
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
 Resolving package versions...
 Installed SnoopCompile ──── v0.3.0
 Installed PackageCompiler ─ v0.4.0
 Installed LibExpat ──────── v0.5.0
 Installed BinaryProvider ── v0.4.2
 Installed ArgParse ──────── v0.6.0
 Installed Libz ──────────── v1.0.0
 Installed URIParser ─────── v0.4.0
 Installed Compat ────────── v1.1.0
 Installed TextWrap ──────── v0.3.0
 Installed WinRPM ────────── v0.4.2
 Installed BinDeps ───────── v0.8.10
 Installed BufferedStreams ─ v1.0.0
 Installed HTTPClient ────── v0.2.1
 Installed LibCURL ───────── v0.3.2
  Updating `~/.julia/environments/v1.0/Project.toml`
  [9b87118b] + PackageCompiler v0.4.0
  Updating `~/.julia/environments/v1.0/Manifest.toml`
  [c7e460c6] + ArgParse v0.6.0
  [9e28174c] + BinDeps v0.8.10
  [b99e7846] + BinaryProvider v0.4.2
  [e1450e63] + BufferedStreams v1.0.0
  [34da2185] + Compat v1.1.0
  [0862f596] + HTTPClient v0.2.1
  [b27032c2] + LibCURL v0.3.2
  [522f3ed2] + LibExpat v0.5.0
  [2ec943e9] + Libz v1.0.0
  [9b87118b] + PackageCompiler v0.4.0
[...]

PR #114, opened 2018-09-04
fix compile error for 1.0

bajizhh:

PR #112, opened 2018-09-03
Make PackageCompiler work with Julia 1.0

JaapHaitsma:
With these fixes Package Compiler works for me with Julia 1.0

PR #120, opened 2018-09-18
Add note to start of the README: 0.7 isn't supported yet

NHDaly:
We've been seeing a lot of issues and PRs that 0.7/1.0 aren't working (#109, #111, #112, #113, #114, #115, #119) so i think we should add a note to the README that it's not yet supported, but it's a work in progress.

If you want, we can also link to the actual WIP PRs in the README, which might be a good idea as well.


lucatrv:
Good idea!

PR #122, opened 2018-09-20
shared_init option to export jl_runtime initialization into shared lib

phlavenk:
A shared lib (DLL) produced by static compilation still requires some non-trivial housekeeping in order to safely use the `base.@ccallable` exported functions. Not that terrible when dealing with C++, but a very unsatisfying experience when dealing with C# (as in my case).
Therefore I've integrated into the shared library build a minimal C program that implements init_jl_runtime and exit_jl_runtime functions, so in order to use the shared library one doesn't need anything else but these two functions and his own exported functions.

### DIFF
--- a/juliac.jl
+++ b/juliac.jl
@@ -44,6 +44,9 @@ Base.@ccallable function julia_main(args::Vector{String})::Cint
         "--shared", "-s"
             action = :store_true
             help = "build shared library"
+        "--shared-init", "-i"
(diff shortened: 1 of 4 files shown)


lucatrv:
Merged with PR #123, thanks!

lucatrv:
@vtjnash  I saw you have already fixed all this in PR #116, thanks!

PR #143, opened 2018-10-29
Fix a typo: JULIAC_PROGRAMLIBNAME

tkf:
Otherwise compiling with `--init-shared` fails so I guess it is a typo?

BTW, I needed to pass the full path to .so file to `jl_init_with_image` for this example to work: https://discourse.julialang.org/t/aot-compiling-using-packagecompiler/16911/5  It would be nice if `init_jl_runtime` handles such case appropriately (e.g., locate the path using dladdr?).


musm:
LGTM

lucatrv:
Thanks!

PR #158, opened 2018-12-15
snooping with pkg3

SimonDanisch:


src/snooping.jl
```diff
@@ -1,94 +1,156 @@
...
+    Symbol.(keys(project_deps))
+end
+
+"""
+Recursively get's all dependencies for `package` and makes sure they're installed!
+Note, that this will mutate your current Pkg environment!
+Returns all dependencies!
+"""
+function recursive_install_dependencies(package::Symbol, deps = Set{Symbol}(), visited = Set{Symbol}(); installed = Pkg.installed())
+  package in visited && return deps
```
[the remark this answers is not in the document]
SimonDanisch: That doesn't install anything recursively - but all recursive dependencies need to be available to get the most of the generated precompile statements. Or is there an option to do that?

src/snooping.jl
```diff
@@ -1,94 +1,156 @@
...
         return precompile_file
     end
     # merge all of the temporary files into a single output
     open(userimg, "w") do output
-        println(output, """
-            # Prevent this from being put into the Main namespace
-            Core.eval(Module(), quote
-            """)
+        # Prevent this from being put into the Main namespace
+        println(output, "module CompilationModule")
```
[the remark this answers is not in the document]
SimonDanisch: Yes, this is still WIP... eval was segfaulting with 19.000 statements and I couldn't figure out why, so here we are ;)

src/snooping.jl
```diff
@@ -1,94 +1,156 @@
...
+  end
+  foreach(x-> recursive_install_dependencies(x, deps, visited; installed = installed), copy(deps))
+  deps
+end
+
+
+function snoop(package, snoopfile, outputfile; additional_packages = Symbol[])
+    command = """
+    using Pkg, $package
+    using $(join(additional_packages, ", "))
```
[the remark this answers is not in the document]
SimonDanisch: Because I don't have a toml with all recursive deps. But I admit, a cleaner fix would be to generated a toml with all recursive deps and activate that! At this early stage, I was just trying to quickly get things running to see if it even works. If you have some time, it would be great if you can help with the Pkg part of this, since I'm sure your solutions will be much leaner ;) I'm still in a state, where I'm not a 100% sure how to nicely put together a full Pkg3 solution!

SimonDanisch:
I just realized, that maybe the cleanest would be to have a new function for Pkg3 packages a la:

`PackageCompiler.compile("path/to/toml")`, which contains the packages you want to compile.
We can then generate a new toml with all the recursive dependencies and compile from there! The question is then, how to cleanly supply the precompile files.

KristofferC:
You most likely want to have a Manifest (so you know what you are compiling) and that already contains all recursive dependents.

If you don't have a manifest, `Pkg.instantiate` creates one for you.

src/snooping.jl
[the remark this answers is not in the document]
pfitzseb: How about generating two files, one with only the precompile statements and the other with
```
Core.eval(Module(), :(include("the_other_file.jl"))
```
or just calling that line directly instead of including a file?

SimonDanisch:
yeah but all recursive deps need to be _installed_ so that you can say using. that wouldn't be the case with activate, right?

SimonDanisch:
or can i actually "install" all packages in a manifest, without a project toml that mentions them all?

pfitzseb:
AFAIU they are all installed, you just can't load them from `Main`. You could potentially generate the manifest with `instantiate` and then `add` all packages in there. Or prefix all modules properly, I guess.

KristofferC:
Ok so the problem is that the package need to be loaded for the precompile statements to work.

It is possible to load a package from its uuid directly using `Base.require`. We can then just do something like `const Package = Base.require(Base.PkgId("Package", PACKAGE_UUID)))`. That way packages in the manifest but not in the project can still be loaded and precompiled.

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+
+"""
+    run_julia(
+        code::String;
+        g = nothing, O = nothing, output_o = nothing, J = nothing,
+        startup_file = "no"
+    )
+
+Runs `code` in a new julia command!
+You can overwrite any julia command line flag by setting it to a value.
```
KristofferC: Is this really true? Doesn't look like `julia_code_cmd` accepts any kw. Or do you mean just the `g, O` etc flags?

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+    add_command!(cmd.exec, false, jl_cmd, "-O", O)
+    add_command!(cmd.exec, false, jl_cmd, "-J", J)
+
+    add_command!(cmd.exec, true, jl_cmd, "-g", g)
+    add_command!(cmd.exec, true, jl_cmd, "--output-o", output_o)
+    add_command!(cmd.exec, true, jl_cmd, "--startup-file", startup_file)
+    add_command!(cmd.exec, true, jl_cmd, "--compile", compile)
+    add_command!(cmd.exec, true, jl_cmd, "--trace-compile", trace_compile)
+    # for better debug, let's not make a tmp file and immediately delete it!
+    file = sysimg_folder("run_julia_code.jl")
```
KristofferC: Is this created somewhere?

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+        toml_path::String, snoopfile::String;
+        force = false, reuse = false, verbose = true,
+        debug = false, cc_flags = nothing
+    )
+
+    Extract all calls from `snoopfile` and ahead of time compiles them
+    incrementally into the current system image.
+    `force = true` will replace the old system image with the new one.
+    The `toml` will need to contain all packages that `snoopfile` uses explicitely.
+    Implicitely used packages & modules doen't need to be contained!
```
KristofferC:
```suggestion
    Implicitly used packages & modules don't need to be contained!
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+"""
+function PrecompileCommand(path)
+    InitBase() * Fix30479() * InitREPL() * Include(path)
+end
+
+
+"""
+    extract_flag(flag, jl_cmd = Base.julia_cmd())
+
+Extracts the value for `flag` from the current julia cmd
```
KristofferC:
```suggestion
Extracts the value for `flag` from `jl_cmd`.
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+function extract_flag_value(flag, jl_cmd = Base.julia_cmd())
+    val = extract_flag(flag, jl_cmd)
+    replace(val, flag => "")
+end
+current_systemimage() = extract_flag_value("-J")
+
+"""
+    add_command!(cmd, ignore, jl_cmd, flag, value)
+
+Adds a command to `cmd` - either choosing the julia default for value == nothing,
```
KristofferC:
```suggestion
Add a command to `cmd` - either choosing the julia default if `value == nothing`,
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+end
+
+"""
+    run_julia(
+        code::String;
+        g = nothing, O = nothing, output_o = nothing, J = nothing,
+        startup_file = "no"
+    )
+
+Runs `code` in a new julia command!
```
KristofferC:
```suggestion
Runs `code` in a new julia command!
```
```suggestion
Run `code` in a julia command.
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+You can overwrite any julia command line flag by setting it to a value.
+If nothing is chosen, it will default to the value of the current julia process.
```
KristofferC:
```suggestion
If the flag has the value `nothing`, the value of the flag of the current julia process is used.
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+        debug = false, cc_flags = nothing
+    )
+
+    Extract all calls from `snoopfile` and ahead of time compiles them
+    incrementally into the current system image.
+    `force = true` will replace the old system image with the new one.
+    The `toml` will need to contain all packages that `snoopfile` uses explicitely.
+    Implicitely used packages & modules doen't need to be contained!
+
+    For a simpler version to just compile a single package have a look at:
```
KristofferC:
```suggestion
    To compile just a single package, see the simpler version  `compile_incremental(package::Symbol)`:
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+    `compile_incremental(package::Symbol)`
```
KristofferC:
```suggestion
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+    compile_incremental(
+        toml_path::String, snoopfile::String;
+        force = false, reuse = false, verbose = true,
+        debug = false, cc_flags = nothing
+    )
+
+    Extract all calls from `snoopfile` and ahead of time compiles them
+    incrementally into the current system image.
+    `force = true` will replace the old system image with the new one.
+    The `toml` will need to contain all packages that `snoopfile` uses explicitely.
```
KristofferC:
```suggestion
    The argument `toml_path` should contain a project file of the packages that `snoopfile` explicitly uses.
```

Should there be a manifest too?

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+    # for better debug, let's not make a tmp file and immediately delete it!
+    file = sysimg_folder("run_julia_code.jl")
+    open(io-> println(io, code), file, "w")
+    push!(cmd.exec, file)
+    cmd
+end
+
+
+"""
+    compile_incremental(
```
KristofferC: What does `cc_flags` and `reuse` do?

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+end
+
+"""
+    compile_incremental(
+        package::Symbol;
+        force = false, reuse = false, verbose = true,
+        debug = false, cc_flags = nothing
+    )
+
+    Compiles `package` incrementally into the current system image.
```
KristofferC:
```suggestion
    Incrementally compile `package` into the current system image.
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+    `force = true` will replace the old system image with the new one.
+    compile_incremental will automatically use the Package/test/runtests.jl to
```
KristofferC:
```suggestion
    `compile_incremental` will run the `Package/test/runtests.jl` file to
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+    figure out what functions to compile. So the coverage of Package's tests will
```
KristofferC:
```suggestion
    and record the functions getting compiled. The coverage of the Package's tests will
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+    directly correlate with how much is getting ahead of time compiled.
```
KristofferC:
```suggestion
    thus determine what is getting ahead of time compiled.
```

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
...
+    For a less magical version of compile_incremental, have a look at:
```
KristofferC:
```suggestion
    For a more explicit version of compile_incremental, see:
```

src/pkg.jl
```diff
@@ -0,0 +1,218 @@
+using Pkg
+using Pkg: TOML
+
+const manifest_memoize = Ref{Dict{String, Any}}()
+
+function current_manifest()
+    if !isassigned(manifest_memoize)
+        manifest_memoize[] = TOML.parsefile(replace(Base.active_project(), "Project.toml" => "Manifest.toml"))
```
KristofferC: Project files can be called `JuliaProject.toml` as well.

KristofferC: Can also use `Pkg.Types.Context().env.manifest`

src/incremental.jl
SimonDanisch: Yeah, it's not fully implemented yet ;) Also it should use the cmd struct that stores the actual flags, even if julia wasn't explicitely started with them. Forgot where to find that struct, so I implemented it with julia_cmd for now

src/incremental.jl
SimonDanisch: The folder? Should be, it's the folder I store stuff related to system image creation.

src/incremental.jl
SimonDanisch: cc_flags is for the linker as far as I'm concerned... I didn't actually add that flag, so I'm not sure if it's even needed for this use case, but I thought why not pass it through when compile_shared accepts it.

src/incremental.jl
SimonDanisch: Good question. Isn't that used implicitly by activate when stored in the same folder?

src/pkg.jl
SimonDanisch: That returns a dict with quite a different format... I think I'll keep it this way for now, and we can make Pkg improvements in another PR ;) I'm sure you have quite a few more improvements!

src/incremental.jl
```diff
@@ -0,0 +1,183 @@
+
+"""
+Init basic C libraries
+"""
+function InitBase()
```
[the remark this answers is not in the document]
SimonDanisch: I first liked the idea of having different types/modules with documentation for what each component does. Since I'd need to make a type for it and define `*` on them, I thought it's much simpler to just use a function and return a string - but I kept the capitalization from that idea.

SimonDanisch:
I tested compilation of quite a few packages and it seems to work relatively well.
Hence, I plan to  merge & tag today!

There is still lots to do, but I'd prefer to do these things in new PRs:

* rework Pkg integration
* better interfaces
* deprecate old compile_package methods, since they're severely inferior now
* add an option for non incremental compilation
* refactor static_julia to clean up the whole positional argument mess 
* unify `run_julia` & `build_julia_cmd` ... `run_julia` + `julia_code_cmd` should actually build upon a method like `run_file`

PR #166, opened 2019-01-22
Display the broken julia_cmd on error so the user can identify the problem

NHDaly:
It looks like PackageCompiler isn't compatible with Julia `v1.1.0`: the command-line args in `Base.julia_cmd()` have changed.

This PR is a tiny one that just helps the user see what's wrong in the output:

```julia
ERROR: LoadError: Unexpected format of "Base.julia_cmd()", you may be using an incompatible version of Julia:
`/Users/nathan.daly/src/julia_release_native/usr/bin/julia -Cnative -J/Users/nathan.daly/src/julia_release_native/usr/lib/julia/sys.dylib --check-bounds=yes -g1`
```

I'll follow up with an actual fix to the above v1.1 problem! :)

### DIFF
--- a/src/static_julia.jl
+++ b/src/static_julia.jl
@@ -106,7 +106,7 @@ function static_julia(
         isfile(juliaprog) || error("Cannot find file: \"$juliaprog\"")
         quiet || println("Julia program file:\n  \"$juliaprog\"")
     end
-    
+
     if executable
         cprog = abspath(cprog)
         isfile(cprog) || error("Cannot find file: \"$cprog\"")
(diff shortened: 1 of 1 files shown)


NHDaly:
Actually, i don't really see why we have this check at all... We may as well give it a shot, no? I don't see why the user shouldn't be allowed to start julia with custom flags and have those carry-through when compiling (e.g. like `--math-mode=fast` or something)

NHDaly:
Ah, jk i get it, it's cause we want to assert that the structure matches what we expect so we can deal with it in the following line. :)

K nvm, I think this PR should be merged as is, and i'll work on a fix for `v1.1` in a follow-up! :)

NHDaly:
(The build failures are for `nightly`, which of course we know is broken. But the good news is now at least the errors include the context:
```julia
ERROR: LoadError: Unexpected format of "Base.julia_cmd()", you may be using an incompatible version of Julia:
`/home/travis/julia/bin/julia -Cnative -J/home/travis/julia/lib/julia/sys.so --check-bounds=yes -g1 --code-coverage=user`
```
https://travis-ci.org/JuliaLang/PackageCompiler.jl/jobs/483001973)

lucatrv:
Yes I confirm that I put that check to assert that the structure matches what we expect.

PR #167, opened 2019-01-22
Fix for julia `v1.1`: allow `Base.julia_cmd()` to contain unexpected flags (adds future-compatibility)

### LINKED ISSUE #162: Unexpected format of "Base.julia_cmd()" compatability issues with 1.2.0-dev.60
Version: `1.2.0-dev.60`

`ERROR: Unexpected format of "Base.julia_cmd()", you may be using an incompatible version of Julia`
is thrown when calling: `PackageCompiler.compile_package("Makie", force=false)`

help in REPL returned this:
```
help?> Base.julia_cmd()
  Base.julia_cmd(juliapath=joinpath(Sys.BINDIR::String, julia_exename()))

return a julia command similar to the one of the running process. 
Propagates any of the --cpu-target, --sysimage, --compile,
 --sysimage-native-code, --compiled-modules, --inline, --check-bounds, --optimize, -g, --code-coverage, and --depwarn command
  line arguments that are not at their default values.

 Julia 1.1
  │
  │  Only the --cpu-target, --sysimage, --depwarn, --compile and --check-bounds flags were propagated before Julia 1.1.
```
when `Base.julia_cmd()` is called it only returns 4 arguments, however 5 are expected. 

at line 213 in static_julia.jl:
```
    julia_cmd = `$(Base.julia_cmd())`
    if length(julia_cmd.exec) != 5 || !all(startswith.(julia_cmd.exec[2:5], ["-C", "-J", "--compile", "--depwarn"]))
        error("Unexpected format of \"Base.julia_cmd()\", you may be using an incompatible version of Julia")

```
Merry Christmas.

NHDaly:
Follow up to https://github.com/JuliaLang/PackageCompiler.jl/pull/166.

Search through the flags in `Base.julia_cmd()` to find the ones we want to edit, rather than expecting them to be in a certain order. This allows future versions of julia to change command line flags without this package breaking.

Fixes #162.

### DIFF
--- a/src/static_julia.jl
+++ b/src/static_julia.jl
@@ -106,7 +106,7 @@ function static_julia(
         isfile(juliaprog) || error("Cannot find file: \"$juliaprog\"")
         quiet || println("Julia program file:\n  \"$juliaprog\"")
     end
-    
+
     if executable
         cprog = abspath(cprog)
         isfile(cprog) || error("Cannot find file: \"$cprog\"")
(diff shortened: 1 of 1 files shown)


NHDaly:
This now failed because now we're passing on the `--code-coverage=user` flag to the julia binary, which seems to be new to julia `v1.1`, and is incompatible with `--output-o`.

I've just pushed up a commit to always set `--code-coverage=no` in the spawned process.

Issue #192, opened 2019-02-23
Modules' __init__ should be called in imported order


tkf:
Currently, `compile_incremental(:PyCall)` generates `incremental_precompile.jl` containing something like

```julia
using Pkg, Statistics, Test, Libdl, Conda, Serialization, LinearAlgebra,
    PyCall, VersionParsing, MacroTools, Dates, PackageCompiler
for Mod in [Pkg, Statistics, Test, Libdl, Conda, Serialization, LinearAlgebra,
    PyCall, VersionParsing, MacroTools, Dates, PackageCompiler]
    isdefined(Mod, :__init__) && Mod.__init__()
end
```

`PyCall` depends on `MacroTools` (`Conda`, and `VersionParsing`).  Furthermore, in principle, `PyCall`'s top-level code and macros can depend on the global state configured in `MacroTools.__init__` (`PyCall` does not depend on it ATM. However, you can check it by adding, e.g., `println(MacroTools.@expand @time foo())` in `PyCall.__init__`).  So, shouldn't `using MacroTools` and `MacroTools.__init__()` come before `using PyCall`?  That is to say, I think `incremental_precompile.jl` needs to contain something like

```julia
for name in [...,
            :MacroTools,
            :VersionParsing,
            :Conda,
            :PyCall,
            ...]
    Core.eval(@__MODULE__, :(using $name))
    Mod = Core.eval(@__MODULE__, name)
    isdefined(Mod, :__init__) && Mod.__init__()
end
```

In general, I think the packages have to be imported and `__init__`'ed in the topologically sorted order.

tkf:
You can find a function that does this here: https://github.com/tkf/julia-sysimage-recipes/blob/master/lib/moduleloader.jl

KristofferC:
Why are `__init__` getting called explicitly at all?

tkf:
Please see https://github.com/JuliaLang/PackageCompiler.jl/issues/277#issuecomment-543516557

PR #204, opened 2019-03-19
remove complex toml based snooping

SimonDanisch:


src/snooping.jl
```diff
@@ -1,23 +1,44 @@
...
-        Pkg.activate($(repr(tomlpath)))
-        Pkg.instantiate()
-        """
-    end
+(adds it only for the current session).
+"""
+function init_package!(packages::Symbol...)
+    push!(packages_needing_initialization, packages...)
+    # TODO use set?!
+    unique!(packages_needing_initialization)
```
tkf: I'm not sure how `init_package!` would be used, but it would be nice if it can track the order that the packages are initialized (so `Set` is out).  I think it is required when the top-level code of downstream packages needs global states configured by `__init__` of the upstream packages.  See: #192

SimonDanisch: good point! It seems like you wrote a method to sort packages topological, would you mind adding that? Although, in this PR I'm not initializing packages by default anymore... You need to do it explicitly by calling `init_package!(mypackages...)` so in theory the user could do that. If we add a toposort method, we could protect against user error though ;)

tkf: I just found an internal function in Pkg that I can use so I didn't need to write anything complicated.  But my code has some problems:

* This internal function is changed between Julia 1.0 and 1.1.  So my code won't work as-is for Julia 1.0.  (But ultimately we can just "vendor" the related code from Pkg so it's an easy problem.)
* The hard part is that being in Manifest.toml does not necessary mean that it should be in the system image.

    * Project.toml/REQUIRE info may be old.  Packages that work only for Julia 0.6 may be listed as dependency even though they are not imported anymore.
    * Some packages are appropriate only for some OSes.  They are usually conditionally imported by something like `if Sys.isapple(); using ...; end`.  This is no good way to avoid importing those packages.

> not initializing packages by default anymore

This is a hard problem but I think a slightly less worse option for now is to initialize all packages by default.  IIUC, we need to be able to evaluate the all top-level statements in the `--output-o` mode, right?  Those top-level statements may require the runtime of the upstream packages.  For example, it's conceivable that someone wants to compute a constant or implement a macro using PyCall functions.  However, PyCall needs to setup Python runtime in its `__init__`.

I think it would be nice if Julia in `--output-o` mode uses precompilation cache.  This way, we don't need to evaluate any code so we can forget about the `__init__` issue.  I bring up this in https://discourse.julialang.org/t/22956/49 as well.  I'm wondering if it is possible to get more attention to this issue...

chriscoey:
Is this branch in a useable state?

lkapelevich:
Possibly related to https://github.com/JuliaLang/PackageCompiler.jl/issues/227 and/or https://github.com/JuliaLang/PackageCompiler.jl/issues/235 
I have a script that reads JLDs which I would like to use with `precompile_incremental`, but I run into errors when loading types from other packages. MWE: https://gist.github.com/lkapelevich/1ef3f7b2e784d5e257c24761bbe6aeac

SimonDanisch:
>Is this branch in a useable state?

Yeah, just needs some clean up, and needs to work with unregistered packages...

lkapelevich:
I also tried using Julia's Serialization- which works for the MWE above. However it gives other troubles when using the deserialized sparse matrices: https://gist.github.com/lkapelevich/f44fceda3f73bc57540bd7182b7e2ae0

SimonDanisch:
Yeah, that may be a Julia bug, or something isn't set up correctly... The serialization code uses lots of internal magic, so it's not entirely surprising that it fails in snoop mode, which changes how code is a run a bit. Only chance to fix this is to further reduce the example, e.g. by diving into the serialization code and finding the exact calls that behave differently.

SimonDanisch:
Nevermind, @lkapelevich issue got fixed by snooping first, and the compiling from the snoop file!

SimonDanisch:
Which can one can do like this:
```julia
using PackageCompiler
open("precompiles.jl", "w") do io
    PackageCompiler.snoop(abspath("the_file_to_snoop.jl"), io)
end
PackageCompiler.compile_incremental(abspath("precompiles.jl"))
```

lkapelevich:
I'm using `SnoopCoompile` to check if anything is recompiling- and it seems that it is although I could be doing something wrong... I edited the [gist ](https://gist.github.com/lkapelevich/1ef3f7b2e784d5e257c24761bbe6aeac) with my code, I may be using something wrong

src/incremental.jl
```diff
@@ -144,35 +144,36 @@ function compile_incremental(
...
+This process requires a script that julia will run in order to determine
+which functions to compile. A package may define a script called `snoopfile.jl`
+for this purpose. If this file cannot be found the package's test script
+`Package/test/runtests.jl` will be used. `compile_incremental` will search
+for `snoopfile.jl` in the package's root directory and in the folders
+`Package/src` and `Package/snoop`. For a more explicit version of compile_incremental,
+see: `compile_incremental(toml_path::String, snoopfile::String)`
+
+Not all packages can currently be compiled into the system image. By default,
+`compile_incremental(:Package) will also compile all of Package's dependencies.
```
cmcaine:
```suggestion
`compile_incremental(:Package)` will also compile all of Package's dependencies.
```

Missing a backtick here.

src/pkg.jl
```diff
@@ -101,13 +101,24 @@ const relative_snoop_locations = (
...
     idx === nothing && error("No snoopfile or testfile found for package $pkg_root")
     return paths[idx]
 end
 
+"""
+    snoop2root(snoopfile) -> pkg_root
+Given a path to a snoopfile, this function returns the package root.
+If the file isn't of a known format it will return nothing.
+reverse of get_snoopfile(pkg_root)
+"""
```
GunnarFarneback: I don't think the implementation matches this.
```
julia> PackageCompiler.snoop2root("package/snoopfile.jl")
"package/"

julia> PackageCompiler.snoop2root("package/snoop/snoopfile.jl")
"package/snoop/"

julia> PackageCompiler.snoop2root("package/test/snoopfile.jl")
"package/test/"

julia> PackageCompiler.snoop2root("package/test/runtests.jl")
"package/"
```
Seems a bit arbitrary so I guess the implementation is off.

src/pkg.jl
```diff
@@ -94,18 +94,27 @@ end
 
 get_snoopfile(pkg::Types.PackageSpec) = get_snoopfile(root_path(pkg))
 
+const relative_snoop_locations = (
+    "snoopfile.jl",
+    joinpath("snoop", "snoopfile.jl"),
+    joinpath("test", "snoopfile.jl"),
+    joinpath("test", "runtests.jl"),
+)
```
GunnarFarneback: This list seems a little inflexible. I wanted to collect my snoopfile and a script to call PackageCompiler in a directory of my choice, say `compile`. If `relative_snoop_locations` had been a vector instead of a tuple I could have modified it as needed, but maybe there should be a more direct API to specify such things.

SimonDanisch: Creating your own non package specific snoop files is a pretty different task so will need a different api (or at least a different function to call).
E.g. resolving packages etc will need to work much differently

GunnarFarneback: Really? I made a shortened version of my `test/runtests.jl` and it seemed to work fine as a snoop file. It was very much package specific, just that I wanted the naming differently.

src/incremental.jl
```diff
@@ -127,7 +127,7 @@ end
     To compile just a single package, see the simpler version  `compile_incremental(package::Symbol)`:
 """
 function compile_incremental(
-        toml_path::Union{String, Nothing}, precompiles::String;
```
tkoolen: What is the motivation for removing the `toml_path` argument here? It's currently unused (#222), but why shouldn't we just fix that (#259) rather than removing it? I'm using this functionality.

src/snooping.jl
```diff
@@ -1,23 +1,46 @@
 using Pkg, Serialization
 
 
-function snoop(package, tomlpath, snoopfile, outputfile, reuse = false, blacklist = [])
```
tkoolen: I'm using this function. What is the motivation for removing it? What will be the upgrade path?

arnavs:
The two tests that fail on this PR (lines `25` and `38` of `runtests.jl`) are because of this function: 

```
function snoop_packages(
        packages::Vector{String}, file::String;
        install::Bool = false, verbose::Bool = false
    )
    ctx = Pkg.Types.Context()
    pkgs = PackageSpec.(packages)
    snoopfiles = get_snoopfile.(pkgs)
    packages = flat_deps(ctx, packages)
    direct_test_deps = test_dependencies(pkgs)
    missing_pkgs = not_installed(Types.PackageSpec[direct_test_deps...])
    if install && !isempty(missing_pkgs)
        Pkg.API.add_or_develop(ctx, missing_pkgs, mode = :add)
    else
        @warn("The following test dependencies are not installed: $missing_pkgs.
        Snooping based on test scripts will likely fail.
        Please use `install = true` or install those packages manually")
    end
    # get all recursive test deps:
    union!(packages, flat_deps(ctx, direct_test_deps))
    # remove blacklisted packages from full list of packages
    imports = setdiff(to_pkgid.(packages), resolved_blacklist(ctx))
    inits = intersect(imports, resolved_inits(ctx))
    usings = join(["const $(x.name) = Base.require($(prepr(x)))" for x in imports], "\n")
    inits = join("    " .* getfield.(inits, :name), ",\n")
    open(file, "w") do io
        println(io, """
        $usings
        __init_modules = [
        $inits
        ]

        for Mod in __init_modules
            isdefined(Mod, :__init__) && Mod.__init__()
        end
        """)
        for file in snoopfiles
            snoop(file, io; verbose = verbose)
        end
    end
end
```

Specifically, it identifies the paths to the right `snoopfiles`, but the packages are not actually in the `ctx` context coming out of `Pkg`. My gut is that inside the refactoring, we missed a step where these things are `pkg"added"`, or something like that. 

Does anyone have any advice on this score?

src/incremental.jl
tkoolen: OK, https://github.com/JuliaLang/PackageCompiler.jl/pull/204/commits/e12e89951919df8a82a235143ea4b2a17b9e4eb5 addresses this concern by just using the project of the parent process. Thanks, @SimonDanisch.

src/snooping.jl
tkoolen: Also addressed in https://github.com/JuliaLang/PackageCompiler.jl/pull/204/commits/e12e89951919df8a82a235143ea4b2a17b9e4eb5 as far as I can tell.

asinghvi17:
2 problems I'm getting here:
a) using `install=true` in `compile_incremental` fails, with the error `add_or_develop not defined`, which is caused by https://github.com/JuliaLang/PackageCompiler.jl/blob/e12e89951919df8a82a235143ea4b2a17b9e4eb5/src/snooping.jl#L182.  Fixing this (by renaming `add_or_develop` to `add`, and removing the `mode` argument fixes the error.
b) The warning:
```
┌ Warning: The following test dependencies are not installed: Pkg.Types.PackageSpec[].
│         Snooping based on test scripts will likely fail.
│         Please use `install = true` or install those packages manually
└ @ PackageCompiler ~/.julia/packages/PackageCompiler/sTrwT/src/snooping.jl:184
```
is thrown no matter what.  Perhaps that should be kept inside an `if` statement?

KristofferC:
Obsolete by https://github.com/JuliaLang/PackageCompiler.jl/pull/304 but I guess I shouldn't close this PR we probably want the commits here to stick around since people might have it in their Manifest.

GunnarFarneback:
I would say that's a fair assumption. Personally I rely on a private fork off this branch since I needed some small additional modifications, but there are probably people using this directly. I also guess not everybody (including me) is ready to drop support for older Julia versions than 1.3.1 in order to use the new code.

KristofferC:
As long as the branch is not deleted, the commit should still be here though.

DilumAluthge:
Now that PackageCompiler has been completely rewritten, should this PR be closed?

SimonDanisch:
I guess we can close it and leave the branch

PR #210, opened 2019-03-25
Fix Regression in revert()

arnavs:
@SimonDanisch I replace what we had in `revert()` (rolling our own system image using the `compile_system_image`), with the default functions coming out of the Julia `DATAROOTDIR`.

If this looks good to you (it seems to fix the regression on my machine), then I'll update the docs/propagation of the now-defunct `debug` argument. 

If you want to stick with the homegrown stuff, I can try something with `PrecompileCommand`.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -44,6 +44,7 @@ Reverts a forced compilation of the system image. Calls out to
 `Main.BuildSysImg.build_sysimg(force = true)`.
 """
 function revert(debug = false)
+    include(joinpath(Sys.BINDIR, Base.DATAROOTDIR, "julia", "build_sysimg.jl"))
     Main.BuildSysImg.build_sysimg(force = true)
 end
 
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -40,14 +40,11 @@ function julia_cpu_target(::Nothing)
 end
 
 """
-Reverts a forced compilation of the system image.
-This will restore any previously backed up system image files, or
-build a new, clean system image.
+Reverts a forced compilation of the system image. Calls out to 
+`Main.BuildSysImg.build_sysimg(force = true)`.
 """
 function revert(debug = false)
-    syspath = default_sysimg_path(debug)
-    sysimg_backup = dirname(get_backup!(debug))
-    copy_system_image(sysimg_backup, syspath)
+    Main.BuildSysImg.build_sysimg(force = true)
 end
 
 function get_root_dir(path)


src/PackageCompiler.jl
```diff
@@ -40,14 +40,12 @@ function julia_cpu_target(::Nothing)
...
-build a new, clean system image.
+Reverts a forced compilation of the system image. Calls out to 
+`Main.BuildSysImg.build_sysimg(force = true)`.
 """
 function revert(debug = false)
-    syspath = default_sysimg_path(debug)
-    sysimg_backup = dirname(get_backup!(debug))
-    copy_system_image(sysimg_backup, syspath)
+    include(joinpath(Sys.BINDIR, Base.DATAROOTDIR, "julia", "build_sysimg.jl"))
+    Main.BuildSysImg.build_sysimg(force = true)
```
SimonDanisch: I think that functionality is deprecated... Did you actually check that it works and improves the timing?

arnavs: Yeah. This stuff is in the Julia 1.1 docs, but I think it's not going to be in Julia 1.2.

arnavs: See here: https://docs.julialang.org/en/v1/devdocs/sysimg/#Main.BuildSysImg.build_sysimg

SimonDanisch: Are you sure that this solves the startup latency? On first sight, I can't see anything, that `build_sysimg.jl` does better than what PackageCompiler was doing...

arnavs: I'll post a new set of benchmarks after lunch.

arnavs: Actually, I don't think this is the best approach anyway. I ran into a cross-platform bug and GitHub issues suggest... switching to `PackageCompiler,`

I'll try the approach with `PrecompileCommand` or something.

SimonDanisch: `PrecompileCommand ` whats that?

arnavs: https://github.com/JuliaLang/PackageCompiler.jl/blob/fa1f43a56ef7f448a4c9af57dedb37174eaa36f8/src/incremental.jl#L53 this thing, figured it would do what we need (precompile things)

SimonDanisch: That's pretty unrelated ;)

arnavs: rip. Will close this then, until I get a better sense of how the different pieces relate.

SimonDanisch: Thanks for the effort :) I took a look as well, but it's not 100% simple...

arnavs: No problem.

PR #213, opened 2019-04-02
Avoid serializing Base.repl_hooks and Base.package_callbacks

tkf:
This does what is already done for `Base.atexit_hooks` to `Base.repl_hooks` and `Base.package_callbacks`.  Before this PR, I could have two `Requires.loadpkg` entries in `Base.package_callbacks`.  I don't know how to invoke problems related to `Base.repl_hooks` in real-world packages yet.  But it seems that we need this treatment based on observations in other hooks.

### DIFF
--- a/src/incremental.jl
+++ b/src/incremental.jl
@@ -47,11 +47,31 @@ function ExitHooksEnd()
     """
 end
 
+function PackageCallbacksStart()
+    """
+    package_callbacks_copy = copy(Base.package_callbacks)
+    empty!(Base.package_callbacks)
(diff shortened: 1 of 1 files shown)


tkf:
I also added `disable_library_threading_hooks` support.  Without this, calling `LinearAlgebra.__init__` when compiling the system image duplicates the hook.

SimonDanisch:
Didn't know about those, thank you!

SimonDanisch:
CI times out, but seems to work fine!

PR #229, opened 2019-04-24
Use `import` instead of `using` in precompile file

tkf:
`using` of a package may introduce a name same as a package to be imported later.  Importing packages with `import` is safer.


SimonDanisch:
Ah yeah, I did that change in https://github.com/JuliaLang/PackageCompiler.jl/pull/204

tkf:
Awesome :)  Can't wait for the next release!

PR #237, opened 2019-05-15
Update build_sysimg to match docstring.

travigd:
The docstring was inconsistent with the implementation. It still doesn't include **all** of the keyword arguments but it's strictly better than it was.


SimonDanisch:
Thanks!

Issue #242, opened 2019-06-07
PkgError while creating temporary environment


asinghvi17:
```julia
compile_incremental(:Makie, :AbstractPlotting)
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
 Resolving package versions...
  Updating `~/.julia/packages/PackageCompiler/oT98U/packages/Makie/Project.toml`
 [...]
  Updating `~/.julia/packages/PackageCompiler/oT98U/packages/Makie/Manifest.toml`
 [...]
 Resolving package versions...
  Updating `~/.julia/packages/PackageCompiler/oT98U/packages/Makie/Project.toml`
  [9b87118b] + PackageCompiler v0.6.3
  Updating `~/.julia/packages/PackageCompiler/oT98U/packages/Makie/Manifest.toml`
[...]
[ Info: activating environment at `~/.julia/packages/PackageCompiler/oT98U/packages/Makie/Project.toml`.
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
ERROR: LoadError: TypeError: in typeassert, expected VersionNumber, got Pkg.Types.VersionSpec
Stacktrace:
 [1] load_urls(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.2/Pkg/src/Operations.jl:438
 [2] #download_source#66 at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.2/Pkg/src/Operations.jl:562 [inlined]
 [3] download_source at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.2/Pkg/src/Operations.jl:561 [inlined]
 [4] #instantiate#81(::Nothing, ::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(Pkg.API.instantiate), ::Pkg.Types.Context) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.2/Pkg/src/API.jl:486
[...]
in expression starting at /Users/anshul/.julia/packages/PackageCompiler/oT98U/sysimg/run_julia_code.jl:6
[...]

PR #211, opened 2019-03-25
Bug fix: Run Pkg.resolve() after modifying Project.toml

NHDaly:
This fixes `compile_package` and `compile_incremental` for packages that have test-only dependencies.

For example, I was seeing this error when I tried to compile `CSV`:
```
julia> PackageCompiler.compile_package("CSV", force=false)
[ Info: Recompiling stale cache file /Users/nathan.daly/.julia/compiled/v1.1/CSV/HHBkp.ji for CSV [336ed68f-0bac-5ca0-87d4-7b16caf5d00b]
[ Info: Registered package CSV, using already given UUID: 336ed68f-0bac-5ca0-87d4-7b16caf5d00b
Project CSV v0.4.3
    Status `~/.julia/dev/PackageCompiler/packages/CSV/Project.toml`
  (empty environment)
 Resolving package versions...
  Updating `~/.julia/dev/PackageCompiler/packages/CSV/Project.toml`
  [324d7699] + CategoricalArrays v0.5.2
  [a93c6f00] + DataFrames v0.17.1
  [9a8bc11e] + DataStreams v0.4.1
  [7d512f48] + InternedStrings v0.7.0
  [e1d29d7a] + Missings v0.4.0
  [ea10d353] + WeakRefStrings v0.5.8
  [ade2ca70] + Dates
  [a63ad114] + Mmap
  Updating `~/.julia/dev/PackageCompiler/packages/CSV/Manifest.toml`
  [b99e7846] + BinaryProvider v0.5.3
  [324d7699] + CategoricalArrays v0.5.2
  [944b1d66] + CodecZlib v0.5.2
  [34da2185] + Compat v2.0.0
  [a93c6f00] + DataFrames v0.17.1
  [9a8bc11e] + DataStreams v0.4.1
  [864edb3b] + DataStructures v0.15.0
  [7d512f48] + InternedStrings v0.7.0
  [82899510] + IteratorInterfaceExtensions v0.1.1
  [e1d29d7a] + Missings v0.4.0
  [bac558e1] + OrderedCollections v1.0.2
  [189a3867] + Reexport v0.2.0
  [ae029012] + Requires v0.5.2
  [a2af1166] + SortingAlgorithms v0.3.1
  [2913bbd2] + StatsBase v0.29.0
  [3783bdb8] + TableTraits v0.4.1
  [bd369af6] + Tables v0.1.18
  [3bb67fe8] + TranscodingStreams v0.9.3
  [ea10d353] + WeakRefStrings v0.5.8
[...]

PR #246, opened 2019-06-18
Fix for julia 1.2

daviehh:
Fixes https://github.com/JuliaLang/PackageCompiler.jl/issues/242

Running in julia 1.2-rc2, the error is, e.g. `compile_incremental(:Makie)`
> ERROR: LoadError: `Documenter` is a direct dependency, but does not appear in the manifest. If you intend `Documenter` to be a direct dependency, run `Pkg.resolve()` to populate the manifest. Otherwise, remove `Documenter` with `Pkg.rm("Documenter")`. Finally, run `Pkg.instantiate()` again.

As per https://github.com/JuliaLang/Pkg.jl/issues/1106 and https://github.com/JuliaLang/Pkg.jl/pull/1209, 
> run `Pkg.resolve()` to populate the manifest. Finally, run `Pkg.instantiate()` again.

to use direct dependency and get a correct toml file.

### DIFF
--- a/src/snooping.jl
+++ b/src/snooping.jl
@@ -13,6 +13,8 @@ function snoop(package, tomlpath, snoopfile, outputfile, reuse = false, blacklis
         append!(Base.LOAD_PATH, $(repr(Base.LOAD_PATH)))
         Pkg.activate($(repr(tomlpath)))
         Pkg.instantiate()
+        Pkg.resolve()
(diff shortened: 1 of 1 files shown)


daviehh:
oops, duplicate https://github.com/JuliaLang/PackageCompiler.jl/pull/211

PR #266, opened 2019-08-10
Fix broken Pkg API use in sd-notomls

asinghvi17:
Also fix a warning being thrown when unnecessary.

(Note: This is based off of, and being merged into, `sd-notomls`- not master!)

### DIFF
--- a/src/snooping.jl
+++ b/src/snooping.jl
@@ -134,6 +134,7 @@ function snoop(snoopfile::String, output_io::IO; verbose = false)
 end
 
 """
+    to_pkgid(pspec::Pkg.Types.PackageSpec)
 PackageSpec has bad hashing behavior, so we use PkgId in places
 """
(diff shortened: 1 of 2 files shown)


src/snooping.jl
```diff
@@ -179,8 +180,8 @@ function snoop_packages(
     direct_test_deps = test_dependencies(pkgs)
     missing_pkgs = not_installed(Types.PackageSpec[direct_test_deps...])
     if install && !isempty(missing_pkgs)
-        Pkg.API.add_or_develop(ctx, missing_pkgs, mode = :add)
-    else
+        Pkg.API.add(ctx, missing_pkgs)
```
SimonDanisch: Are you sure this works across Julia versions? I'm pretty sure I was using `add_or_develop` because `add` alone didn't work...

asinghvi17: Possibly it doesn't, let me install Julia 1.0 and check...it definitely fails on 1.3, though.

PR #273, opened 2019-08-28
add note about packages needing explicit init

asinghvi17:


SimonDanisch:
You should add this to sd-notomls, since on master, everything gets inited (at least if you snoop)...

KristofferC:
Obsolete by https://github.com/JuliaLang/PackageCompiler.jl/pull/304. You shouldn't need to explicitly call `__init__`.

Issue #271, opened 2019-08-23
Error on Julia 1.2: ERROR: LoadError: `FixedPointNumbers` is a direct dependency, but does not appear in the manifest.


Alexander-Barth:
When I use PackageCompiler v0.6.4, on Julia 1.2 (Linux 64-bit), I get the error below.
The package FixedPointNumbers is installed in v0.5.3. The same code seem to work on Julia 1.1.1 (tested on a different machine).

```
julia> using PackageCompiler
julia> sysimage = PackageCompiler.compile_package("FixedPointNumbers", verbose = true)
[ Info: Registered package FixedPointNumbers, using already given UUID: 53c48c17-4a7d-5ca2-90c5-79b7896eea93
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
 Resolving package versions...
  Updating `~/.julia/packages/PackageCompiler/CJQcs/packages/FixedPointNumbers/Project.toml`
 [no changes]
 Resolving package versions...
  Updating `~/.julia/packages/PackageCompiler/CJQcs/packages/FixedPointNumbers/Project.toml`
  [9b87118b] + PackageCompiler v0.6.4
  [44cfe95a] + Pkg 
  Updating `~/.julia/packages/PackageCompiler/CJQcs/packages/FixedPointNumbers/Manifest.toml`
  [9e28174c] + BinDeps v0.8.10
  [b99e7846] + BinaryProvider v0.5.6
  [e1450e63] + BufferedStreams v1.0.0
  [34da2185] + Compat v2.1.0
  [0862f596] + HTTPClient v0.2.1
  [b27032c2] + LibCURL v0.5.2
  [522f3ed2] + LibExpat v0.5.0
  [2ec943e9] + Libz v1.0.0
  [9b87118b] + PackageCompiler v0.6.4
  [30578b45] + URIParser v0.4.0
  [c17dfb99] + WinRPM v0.4.2
  [2a0f44e3] + Base64 
  [ade2ca70] + Dates 
  [8bb1440f] + DelimitedFiles 
  [8ba89e20] + Distributed 
  [b77e0a4c] + InteractiveUtils 
  [76f85450] + LibGit2 
  [8f399da3] + Libdl 
  [37e2e46d] + LinearAlgebra 
  [56ddb016] + Logging 
  [d6f4376e] + Markdown 
  [a63ad114] + Mmap 
  [44cfe95a] + Pkg 
[...]

PR #275, opened 2019-09-19
Allow to precompile Makie by fixing issue #271

j-fu:
The hint in https://github.com/JuliaLang/PackageCompiler.jl/issues/271#issuecomment-532443430
helped me to precompile Makie on a clean Julia 1.2 installation on openSUSE Tumbleweed.
Otherwise I had the same issue  with ParameterizedFunctions in place of FixedPointNumbers.

Jürgen

### DIFF
--- a/src/snooping.jl
+++ b/src/snooping.jl
@@ -12,7 +12,7 @@ function snoop(package, tomlpath, snoopfile, outputfile, reuse = false, blacklis
         # Take LOAD_PATH from parent process
         append!(Base.LOAD_PATH, $(repr(Base.LOAD_PATH)))
         Pkg.activate($(repr(tomlpath)))
-        Pkg.instantiate()
+        Pkg.resolve()
         """
     end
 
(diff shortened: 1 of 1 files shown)


src/snooping.jl
```diff
@@ -12,7 +12,7 @@ function snoop(package, tomlpath, snoopfile, outputfile, reuse = false, blacklis
         # Take LOAD_PATH from parent process
         append!(Base.LOAD_PATH, $(repr(Base.LOAD_PATH)))
         Pkg.activate($(repr(tomlpath)))
-        Pkg.instantiate()
+        Pkg.resolve()
```
asinghvi17: Would it make sense to put an instantiate after this?  

```suggestion
        Pkg.resolve()
        Pkg.instantiate()
```

tamasgal:
This just helped me to fix #287 thanks to a hint from @asinghvi17

ianshmean:
This fix just solved this issue for me. Seems like it should be merged and a new version released (there are a few lingering ```LoadError: `xxx` is a direct dependency, but does not appear in the manifest``` issues.)

aminya:
Agree. This should be merged.

aminya:
@SimonDanisch Could you merge this?

SimonDanisch:
I guess... Although I'd recommend using the sd-notomls branch

PR #284, opened 2019-10-15
Add hashbang to juliac.jl

stanleyjs:
Right now you need to use `julia juliac.jl <args>`.  Simply adding this hashbang should make it so one can symlink the compiler into a directory on their $PATH  `ln juliac.jl -T $some_path_dir`.  They may need to `chmod a+x juliac.jl`.  Then it can be called just as `juliac.jl -vas helloworld.jl`

### DIFF
--- a/juliac.jl
+++ b/juliac.jl
@@ -1,3 +1,5 @@
+#!/usr/bin/env julia
+
 # ArgParse can't be a dependency here,
 # so users need to install it.
 using ArgParse, PackageCompiler


kdheepak:
Does this work on Windows too?

stanleyjs:
I am not sure.  I don't have a windows machine to test. I imagine it would definitely work for anyone using the linux subsystem for windows 10.  I'll check with a colleague.

stanleyjs:
@kdheepak a direct analog for the hashbang when running under native windows (not the linux subsystem) does not appear to exist.

SimonDanisch:
Yeah that makes sense, thank you!

Issue #297, opened 2019-12-28
Build failure on ARM32 (Raspberry Pi 4B) plus proposed fix


wiremoons:
Bug found while compiling a test app on a Raspberry Pi 4B. The same code had previously compiled ok under Ubuntu in an x86 machine.

Bug is caused by additional source code that had been added to fix a problem with compiling for Windows 32bit, based on the comment left at the point of breakage. The comment was helpful however in locating the bug :)

I have provided a pull request with a proposed fix that works ok once applied.

Just for completeness:

The following Raspberry Pi operating system and kernel version is being used
```
Linux pi4mobi 4.19.79-v7l+ #1273 SMP Fri Oct 11 18:23:45 BST 2019 armv7l GNU/Linux

Distributor ID: Raspbian
Description:    Raspbian GNU/Linux 10 (buster)
Release:        10
Codename:       buster
```

The Julia version is the pre-complied version from Julia downloads, as follows:
```
Julia Version 1.3.0
Commit 46ce4d7933 (2019-11-26 06:09 UTC)
Platform Info:
  OS: Linux (arm-linux-gnueabihf)
  CPU: ARMv7 Processor rev 3 (v7l)
  WORD_SIZE: 32
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, cortex-a72)
```

Package versions being used are:
```
(v1.3) pkg> status
    Status `~/.julia/environments/v1.3/Project.toml`
  [c7e460c6] ArgParse v0.6.2
  [9b87118b] PackageCompiler v0.6.5
```

The source code of the test app is:
```
# hello_binary.jl
#

function SayHello()
    println("Hello World!")
end

# Main : required PackageComplier boilerplate startup function
Base.@ccallable function julia_main(ARGS::Vector{String})::Cint
    SayHello()
    return 0
end
```

The build command is below, noting that `juliac` is a bash alias as: `alias juliac='$HOME/.julia/packages/PackageCompiler/4yNnV/juliac.jl'`
```
juliac ./hello_binary.jl -tavern hello_binary
```

The build output including the error and failure message is:
```
Julia program file:
  "/home/simon/projects/code/julia/hello_binary/hello_binary.jl"
C program file:
  "/home/simon/.julia/packages/PackageCompiler/4yNnV/examples/program.c"
Build directory:
  "/home/simon/projects/code/julia/hello_binary/builddir"
Make build directory
Build static library "hello_binary.a":
[...]

PR #298, opened 2019-12-28
Fixed ARM 32bit compile failure - see Issue

wiremoons:
Fix proposed as per information provided in **Issue: #297**

**Summary:** currrently compiling on 32bit ARM architicture is broken. Cause is a 32bit check that is meant only for Windows. Making the check specific to 32 bit Window only, fixed the ARM 32bit compile failure.


KristofferC:
Obsolete by https://github.com/JuliaLang/PackageCompiler.jl/pull/304

PR #302, opened 2020-01-27
Allow WinRPM 1.0

mcabbott:
WinRPM was updated in https://github.com/JuliaPackaging/WinRPM.jl/pull/177 so as not to depend on obsolete HTTPClient.jl, which blocked updates to Compat.jl elsewhere.


SimonDanisch:
Thank you!

PR #304, opened 2020-02-09
Total rewrite of the package (aka make PackageCompilerX the new PackageCompiler)

KristofferC:
This PR completely replaces PackageCompiler.jl with that of PackageCompilerX.jl. The maintainers of this package as well as those of PackageCompilerX.jl feels like this is the best way forward.
A few options were considered:

- Just deprecate this repo, rename PackageCompilerX to something else (AppCompiler.jl). There are some drawbacks to this. Firstly, this package name is already pretty good and it would be a shame to waste it. Secondly, it is a bit confusing to have a lot of references spread out over the internet and then have it be deprecated. Also, it would remove the history of everyone contributing to this package which is a shame.
- Just repoint the URL in the registry to PackageCompilerX. This would cause existing manifests to break so it is a non-starter.

Instead, just putting a complete rewrite on top of the existing history felt like the best choice.

Some improvements / changes in PackageCompilerX.jl vs the current PackageCompiler:

- Support for artifacts in Apps. This allows one to distribute apps 
- More CI, including e.g. ARM support.
- Better defaults for cpu targets, matching those that Julia itself uses
- Proper documentation including some examples, tutorials and developer docs.
- No command line interface, this felt like adding a lot of code for little benefit.
- Uses the Artifact system to get a compiler for Windows instead of the largely deprecated WinRPM.
- Drops pre 1.3.1 Julia support.
- There is no need to annotate the entry points of apps with `Base.@ccallable` since that is done automatically.
- The `julia_main` function no longer take any arguments. Just like in normal Julia code, just access the global `ARGS`.
- An app needs to be structured as a package. (project file and `src/Package.jl` file)

I will go through earlier issues and add here which are either fixed by this PR, no longer relevant due to fixes in Julia etc:

After this PR is mergd, we should transfer over issues PR from PackgeCompilerX to this repo.


SimonDanisch:
Yay :)

DilumAluthge:
@KristofferC Are you planning on making a release as soon as you merge this PR?

If not, it might be better to set the version to `1.0.0-DEV`.

SimonDanisch:
Awesome :) Do we want to try to fix the CI, or is that not really doable, since tests take so long? Could we split up stuff?

SimonDanisch:
>No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

I guess this time out can be upped?

KristofferC:
Strange, it passed on ARM in the previous commit.

DilumAluthge:
Travis is passing.

It looks like the AppVeyor error isn’t actually a test failure, but a problem with the configuration of AppVeyor on this repository.

How about we uninstall AppVeyor from this repo and use Travis for the Windows tests? After all, this PR deletes the .appveyor.yml file, which implies that we aren’t planning on using AppVeyor. 

I think you need to be an organization admin of the JuliaLang organization in order to uninstall the AppVeyor app from this repo. @ViralBShah @StefanKarpinski

LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2019 Kristoffer Carlsson <kristoffer.carlsson@juliacomputing.com>
```
SimonDanisch: Hm, I mean I see that this code is mainly by you, but would it make sense to leave me in here, or do we just say, its in the history?

DilumAluthge: Maybe something like: Kristoffer, Simon, Luca, and other contributors? Just to be comprehensive.

SimonDanisch: That sounds fair!

lucatrv: I would appreciate that, as I put a lot of effort on PackageCompiler.jl development.

SimonDanisch: Definitely ;) That I was alone in the license was wrong to begin with!

ViralBShah: Can someone make a suggestion here so it can just be accepted? Otherwise, we absolutely must add the names. Also let's add `Julia Computing` to the mix please.

DilumAluthge: Yeah I'll make a suggestion

ViralBShah:
I deleted the appveyor webhook.

LICENSE
KristofferC: Sorry, didn't think about this, it was just the default that got generated by PackageCompiler. Let's merge all names in here :)

DilumAluthge:
```suggestion
Copyright (c) 2015-2020 Kristoffer Carlsson, Simon Danisch, Luca Trevisani, Julia Computing, and contributors.
```

DilumAluthge: @ViralBShah

DilumAluthge: Not sure who is able to accept the suggested change. Perhaps either @KristofferC or @SimonDanisch

NHDaly:
Hooray! Thanks for the hard work on this, everyone! 😁 The future is bright!

PR #306, opened 2020-02-10
Add default value (no packages) to create_sysimage.

fredrikekre:


KristofferC:
I wonder if a default value of using all packages would be better. And then you give `[]` if you want no packages. Seems more common.

PR #313, opened 2020-02-11
Run Travis on Julia nightly (but allow Travis to fail on Julia nightly)

DilumAluthge:
This pull request:
- Runs Travis on Julia nightly
- Allows Travis to fail on Julia nightly
- Sets `fast_finish` to `true` so that Travis can return a `success` status even if some jobs are still running, as long as all of the still-running jobs are allowed to fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ os:
   - windows
 julia:
   - 1.3
+  - nightly
 addons:
   apt:
     sources:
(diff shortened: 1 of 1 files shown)


KristofferC:
Needs a rebase

DilumAluthge:
Rebased.

KristofferC:
Will fail until #317 is fixed which I will do.

DilumAluthge:
> Will fail until #317 is fixed which I will do.

Technically, it will pass (green checkmark) because Travis is allowed to fail on nightly.

Do you want me to change it so that Travis is required to pass on nightly? Might be a good idea - would make sure we notice it when CI fails on nightly.

KristofferC:
Oh yeah, nvm then.

PR #325, opened 2020-02-11
Rewrite maximum call in base sysimg.jl to fix error when stdlibs is e…

fredrikekre:
…mpty.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -103,8 +103,13 @@ function rewrite_sysimg_jl_only_needed_stdlibs(stdlibs::Vector{String})
     sysimg_content = read(sysimg_source_path, String)
     # replaces the hardcoded list of stdlibs in sysimg.jl with
     # the stdlibs that is given as argument
-    return replace(sysimg_content,
+    content = replace(sysimg_content,
         r"stdlibs = \[(.*?)\]"s => string("stdlibs = [", join(":" .* stdlibs, ",\n"), "]"))
+    # Also replace a maximum call which fails for empty collections,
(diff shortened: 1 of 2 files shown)


src/PackageCompiler.jl
```diff
@@ -103,8 +103,11 @@ function rewrite_sysimg_jl_only_needed_stdlibs(stdlibs::Vector{String})
     sysimg_content = read(sysimg_source_path, String)
     # replaces the hardcoded list of stdlibs in sysimg.jl with
     # the stdlibs that is given as argument
-    return replace(sysimg_content,
+    content = replace(sysimg_content,
         r"stdlibs = \[(.*?)\]"s => string("stdlibs = [", join(":" .* stdlibs, ",\n"), "]"))
+    # Also replace a maximum call with hardcoded value, see https://github.com/JuliaLang/julia/pull/34727
+    content = replace(content, r"maximum\(textwidth\.\(string\.\(stdlibs\)\)\)" => "14")
```
fredrikekre: Why a regex?

KristofferC:
Maybe add a test that creates a sysimage with zero stdlibs?

fredrikekre:
Also tests https://github.com/JuliaLang/PackageCompiler.jl/commit/c02c4e7fd0a51d2f221140e60d146126616612d4

test/runtests.jl
```diff
@@ -71,4 +71,11 @@ end
             @test occursin("HelloWorld artifact at $(realpath(app_compiled_dir))", app_output)
         end
     end
+    # Test creating an empty sysimage
+    tmp = mktempdir()
```
KristofferC: Perhaps only if `!is_slow_ci`

PR #326, opened 2020-02-11
WIP: Add `download_compiler` function to optionally download compilers

DilumAluthge:
This pull request adds the function `PackageCompiler.download_compiler`.

Running `PackageCompiler.download_compiler()` on 64-bit Linux or macOS will download Clang and set `ENV["JULIA_CC"]` to the path of the downloaded compiler.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,7 @@ addons:
     sources:
       - ubuntu-toolchain-r-test
 before_install:
-  - if [ $USE_ARTIFACTS_COMPILER != "true" ] && [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update; fi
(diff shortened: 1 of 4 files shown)


KristofferC:
Isn't there a jll for Clang? I know there are BB builds of Clang at least so we should use those for sure.

DilumAluthge:
> Isn't there a jll for Clang? I know there are BB builds of Clang at least so we should use those for sure.

I had no idea there was a JLL for Clang - hence my attempt to use the official binaries.

We should definitely use Clang_jll if it exists and works.

Unfortunately, googling `Clang_jll` only brings up this: https://github.com/KristofferC/Clang_jll.jl

And I thought I remembered you saying on Slack that your Clang_jll experiment didn't work?

KristofferC:
It didn't work on Windows :P

DilumAluthge:
Ahhh. I see. Well now we have a working solution on Windows! The trick is to get it working on other platforms.

Did you get your `Clang_jll` working with the `PackageCompiler` test suite on macOS and Linux? I kept getting issues with clang trying (and failing) to find to `libtinfo`.

PR #328, opened 2020-02-13
Fix plots example code block

rob-luke:


DilumAluthge:
I believe that the original is correct.

@KristofferC

rob-luke:
Actually both seem to work. My appologies for adding to the noise.

KristofferC:
Perhaps we should use the more explicit `--sysimage`?

DilumAluthge:
> Perhaps we should use the more explicit `--sysimage`?

Sure, that might help reduce some confusion.

docs/src/examples/plots.md
```diff
@@ -33,7 +33,7 @@ using PackageCompiler
 create_sysimage(:Plots, sysimage_path="sys_plots.so", precompile_execution_file="precompile_plots.jl")
 ```
 
-If we now start Julia with the flag `-Jsys_plots.so` and re-time our previous commands:
+If we now start Julia with the flag `--sysimage sys_plots.so` and re-time our previous commands:
```
KristofferC: There is a use of `-J` later in this file as well.

Issue #295, opened 2019-12-09
PackageCompiler: LLVM ERROR


rssdev10:
Hi, I’m getting the following error while compiling my package:

```
ptrtoint not supported for non-integral pointers
inttoptr not supported for non-integral pointers
ptrtoint not supported for non-integral pointers
inttoptr not supported for non-integral pointers
ptrtoint not supported for non-integral pointers
inttoptr not supported for non-integral pointers
LLVM ERROR: Broken function found, compilation aborted!
ERROR: LoadError: failed process: Process(`/Applications/Julia-1.3.app/Contents/Resources/julia/bin/julia --output-o=run.a --track-allocation=none --code-coverage=none --history-file=yes --inline=yes --math-mode=ieee --project=@. --compile=yes --track-allocation=none --sysimage-native-code=yes --sysimage=/Applications/Julia-1.3.app/Contents/Resources/julia/lib/julia/sys.dylib --compiled-modules=yes --optimize=2 /Users/rss/.julia/packages/PackageCompiler/CJQcs/sysimg/run_julia_code.jl`, ProcessExited(1)) [1]

Stacktrace:
 [1] pipeline_error at ./process.jl:525 [inlined]
 [2] #run#565(::Bool, ::typeof(run), ::Cmd) at ./process.jl:440
 [3] run at ./process.jl:438 [inlined]
 [4] #run_julia#1 at /Users/rss/.julia/packages/PackageCompiler/CJQcs/src/compiler_flags.jl:225 [inlined]
 [5] #run_julia at ./none:0 [inlined]
 [6] (::PackageCompiler.var"#13#14"{Base.Iterators.Pairs{Symbol,Nothing,NTuple{14,Symbol},NamedTuple{(:sysimage, :startup_file, :handle_signals, :sysimage_native_code, :compiled_modules, :depwarn, :warn_overwrite, :compile, :cpu_target, :optimize, :debug_level, :inline, :check_bounds, :math_mode),NTuple{14,Nothing}}},String})() at /Users/rss/.julia/packages/PackageCompiler/CJQcs/src/static_julia.jl:263
[...]

PR #329, opened 2020-02-13
add a blacklist for functions that are known to crash Julia when compiled into sysimage

KristofferC:
Ref https://github.com/JuliaLang/PackageCompiler.jl/issues/295

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -170,6 +170,12 @@ function do_ensurecompiled(project, packages, sysimage)
     return nothing
 end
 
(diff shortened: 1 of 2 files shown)


DilumAluthge:
@KristofferC Two suggestions:
1. Instead of making the blacklist a vector of strings, make it a vector of `RegEx`. That way, you could write a regex that, for example, excludes all of the permutedims statements for all dimensions.
2. Allow the user to override the default value of `precompile_blacklist`.

DilumAluthge:
I have opened #333, which implements my suggestions.

PR #333, opened 2020-02-14
Add a blacklist for precompile statements, and allow users to override the blacklist

DilumAluthge:
Closes #329 
Related to #295 
Related to https://github.com/JuliaLang/julia/issues/31156

---

This is a replacement for #329. This PR adds a blacklist for precompile statements that are known to crash Julia when you try to compile them into a sysimage. The differences between this PR and #329 are:
1. In #329, the user cannot customize the blacklist; it is hardcoded. In this PR, there is a default blacklist, but users are able to customize the blacklist if they choose.
2. In #329, members of the blacklist must be `AbstractString`s. In this PR, members of the blacklist may be either `AbstractString`s or `Regex`es. This allows a user to block any precompile statements that match any pattern in a set of given patterns.

cc: @KristofferC 

@ianshmean may be interested in this, especially the `Regex` part (see his comment here: https://github.com/JuliaLang/PackageCompiler.jl/issues/295#issuecomment-585830915)

---

Just a quick word about the motivation for this PR. I don't think it makes sense for the default blacklist to, for example, blacklist all `permutedims` methods. That is way too aggressive for most users. So we want to have a relatively conservative blacklist. But we want users to be able to provide a custom blacklist as well. And because a user may need to, for example, block methods with `Array{Bool, N}` for all values of `N`, we need to allow them to pass `Regex`es for ultimate customization ability. But users that are not experienced with regexes may instead just want to pass a vector of strings. That is why this PR supports both `Regex`es and `AbstractString`s.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -9,6 +9,12 @@ export create_sysimage, create_app, audit_app, restore_default_sysimage
 
 include("juliaconfig.jl")
 
+const default_precompile_blacklist = [
+    "precompile(Tuple{typeof(Base.permutedims), Array{Bool, 2}, Array{Int64, 1}})",
+    "precompile(Tuple{typeof(Base.permutedims), Array{Bool, 3}, Array{Int64, 1}})",
+    "precompile(Tuple{typeof(Base.permutedims), Array{UInt8, 3}, Array{Int64, 1}})",
+]
+
 const NATIVE_CPU_TARGET = "native"
 # See https://github.com/JuliaCI/julia-buildbot/blob/489ad6dee5f1e8f2ad341397dc15bb4fce436b26/master/inventory.py
 function default_app_cpu_target()
@@ -177,7 +183,8 @@ function create_sysimg_object_file(object_file::String, packages::Vector{String}
                             precompile_statements_file::Vector{String},
                             cpu_target::String,
                             script::Union{Nothing, String},
-                            isapp::Bool)
+                            isapp::Bool,
+                            precompile_blacklist::AbstractVector{<:Union{AbstractString, Regex}} = default_precompile_blacklist)
 
     # Handle precompilation
     precompile_statements = ""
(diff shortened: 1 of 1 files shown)


DilumAluthge:
@ianshmean could you try out this PR and let me know if it solves your issues.

I would recommend using a custom blacklist that looks something like this:
```julia
precompile_blacklist = [r"permutedims"i]
```

This will cause anything with `permutedims` in it to be blacklisted.

ianshmean:
Seems to be working. It helped me get past `permutedims` to see that `copyto!` was causing the same issue, thanks to the LLVM function name report that's on julia master. 
Working my way through. I'll share the list once successful

ianshmean:
I haven't managed to skip the `copyto!` error. It seems to be being invoked by an unknown function.
`[r"permutedims"i, r"copyto!"i]` doesn't filter it out, nether does `r"copy"i`

KristofferC:
Yeah, I am not sure this is a feasible solution since other functions can probably end up calling it. And `copyto!` is pretty common...

DilumAluthge:
@ianshmean What happens if your blacklist is`[r"Array{Bool,"i]`?

DilumAluthge:
I wonder if `copyto!` is being used in the broadcasting machinery.

ianshmean:
I can try that. 

I’ve actually found that my package doesn’t fail on an amd64 Linux machine while it does on the aarch64 system I’ve been testing. I’ve outputted all the precompile statements from both and I’m going to do a symdiff on them to figure out what’s different

ianshmean:
Not so simple..
```
[ Info: Num statements on aarch64: 6406
[ Info: Num statements on amd64: 6096
[ Info: Statements generated on aarch64 but not on amd64: 1537
[ Info: Statements generated on amd64 but not on arm64: 1277
```

aminya:
Isn't better to use `parcel` function from SnoopCompiler? 
cc: @timholy

DilumAluthge:
@ianshmean I think you’re going to have to do a bisect on all of the precompile statements. Probably run it overnight.

KristofferC:
We already kinda did that but AFAIU even if you remove the one that fails, another one can just take its place because running one `precompile()` line compiles a lot of dependent stuff behind the scenes.

DilumAluthge:
@KristofferC What are your thoughts on this PR? It seems like it hasn’t actually helped Ian fix his problem. Should we close this PR and hope for a fix from upstream (Julia and/or LLVM)?

KristofferC:
I think so yes.

PR #335, opened 2020-02-14
WIP: Try to fix debug messages

DilumAluthge:


KristofferC:
Might need https://github.com/JuliaLang/julia/pull/34509.

DilumAluthge:
> Might need [JuliaLang/julia#34509](https://github.com/JuliaLang/julia/pull/34509).

Well, if we set JULIA_DEBUG=all, we should see all of the debug messages even without https://github.com/JuliaLang/julia/pull/34509

KristofferC:
I think we can just wait for 1.4 for this to fix itself?

PR #336, opened 2020-02-14
Suggestion to add `precompile_statements_file_out` option

IanButterworth:
For debugging purposes. Just saves the total list to file.
Not sure if helpful in the broader sense. Just a suggestion.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -276,13 +276,11 @@ function create_sysimg_object_file(object_file::String, packages::Vector{String}
     try
         run(cmd)
     catch e
-        bt = catch_backtrace()
         if length(tracefiles) > 0
-            @error "Julia experienced an error while compiling the sysimage. The precompile statements generated during execution of $precompile_execution_file have been saved to $tracefiles"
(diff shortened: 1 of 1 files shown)


DilumAluthge:
@KristofferC What do you think about merging this? It seems like it would make debugging easier.

KristofferC:
I am hesitant adding too much API and "scaffolding" around https://github.com/JuliaLang/PackageCompiler.jl/issues/295. Also, can't you already get these by just running with `--trace-compile=output_precompile.jl`?

ianshmean:
I think you're right that it's a bit much to add to the API.
How about it dumps the precompile statements to a temp file if LLVM errors? The current error you get is a huge wall of red `Process(...` code, so it might be helpful to make that error nicer at the same time?

KristofferC:
The precompile files used that contain all the statements should be shown in that red wall thought

https://github.com/JuliaLang/PackageCompiler.jl/blob/f82ac2160a8689c17abcafa52cc08b61dacacd9e/src/PackageCompiler.jl#L188-L192

but I agree it might not be very discoverable.

ianshmean:
See what you think of the approach now. I don't feel strongly here, but might be an improvement?

Issue #320, opened 2020-02-11
Multiple executables in an app


KristofferC:
Implementing this is could be done with passing a list of `appname => julia_ccallable_function` or something

KristofferC:
Something like `[(julia_function::String, app_name::String), ...]` might be ok where by default `julia_function = "julia_main"` and `app_name` is taken from the Project.

KristofferC:
We can do this in 2.0 by changing the existing `name` kwarg to `create_app`.

PR #349, opened 2020-02-20
Allow user to specify a custom binary name

kmsquire:
(I need to add documentation.)

### DIFF
--- a/docs/src/apps.md
+++ b/docs/src/apps.md
@@ -172,7 +172,7 @@ app with the resulting sysimage.
 
 By default, the binary in the `bin` directory take the name of the project,
 as defined in `Project.toml`.  If you want to change the name, you can pass
-`binary_name=some_binary_name` to `create_app`.
+`app_name="some_app_name"` to `create_app`.
 
 ### Artifacts
 
@@ -257,4 +257,3 @@ CodeInfo(
 │         value@_3
 │   %11 = MyApp.DEPOT_PATH
 ```
-
--- a/docs/src/apps.md
+++ b/docs/src/apps.md
@@ -168,6 +168,11 @@ multiplication, `rand(3,3) * rand(3,3)` requires both the standard libraries
 Nevertheless, the option is there to use. Just make sure to properly test the
 app with the resulting sysimage.
 
+### Custom binary name
+
+By default, the binary in the `bin` directory take the name of the project,
+as defined in `Project.toml`.  If you want to change the name, you can pass
+`binary_name=some_binary_name` to `create_app`.
 
 ### Artifacts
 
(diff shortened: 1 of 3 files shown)


kmsquire:
Documentation added.

KristofferC:
It would be good to add a test. On the other hand, adding a full app compilation to the test suite feels a bit excessive. Perhaps we can

```
binary_name_args = ("CustomName", nothing)
```

to under

https://github.com/JuliaLang/PackageCompiler.jl/blob/4611a790cb2b19da7a1c91203fdcb4a1bef9015c/test/runtests.jl#L52

and then do

```jl
for (filter, name) in zip(binary_name_args, filter_stdlibs) 
```

instead of

https://github.com/JuliaLang/PackageCompiler.jl/blob/4611a790cb2b19da7a1c91203fdcb4a1bef9015c/test/runtests.jl#L56

(and pass in the argument to `create_app`).

KristofferC:
Maybe, https://github.com/JuliaLang/PackageCompiler.jl/issues/320 is related.

kmsquire:
> Maybe call it `app_name` instead?

My first iteration did this, but I found that I still needed to use `app_name` for the sysimg.  But there's probably a way around this.

I'll try to add a test, and see if there's a nice way around `app_name`.

#320 is also interesting.  

I originally wanted to structure my project with binary files under `bin`, and have them call into `src` (so `src` was essentially a library).  I could have simply made a separate package, but I'm inclined to keep everything together in the same package, so it would be nice to be able to produce multiple binaries.  I might look into that if no one else does, but I'm not sure how soon.

kmsquire:
3 failures:

1. timeout while creating the base system image (Linux Arm64 Julia 1.4)
2. unable to download packages (Linux x86 nightly)
3. missing Windows artifacts (Windows AMD64 nightly)

Everything else passed.

docs/src/apps.md
```diff
@@ -168,6 +168,11 @@ multiplication, `rand(3,3) * rand(3,3)` requires both the standard libraries
 Nevertheless, the option is there to use. Just make sure to properly test the
 app with the resulting sysimage.
 
+### Custom binary name
+
+By default, the binary in the `bin` directory take the name of the project,
+as defined in `Project.toml`.  If you want to change the name, you can pass
+`binary_name=some_binary_name` to `create_app`.
```
KristofferC: This says `binary_name` but elsewhere it says `app_name`.

kmsquire:
```suggestion
`app_name="some_app_name"` to `create_app`.
```

kmsquire: Fixed, thanks.

Issue #360, opened 2020-03-01
syntax error and warning from packagecompiling CodecZlib


jaakkor2:
With `precompile.jl` just one line
```
using CodecZlib
```
prints out
```
julia> create_sysimage([:CodecZlib], sysimage_path="sys_loaders.so", precompile_execution_file="precompile.jl")
[ Info: PackageCompiler: creating system image object file, this might take a while...
C:/Users/jr/.julia/artifacts/572b61b5075459e3ed62317e674398166ca98dd4/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: <unknown-file>:0: syntax error
Warning: .drectve `-export:ccalllib_libz-1.2.11.dll26928,data ' unrecognized
```
with
```
julia> versioninfo()
Julia Version 1.4.0-rc2.0
Commit b99ed72c95 (2020-02-24 16:51 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, skylake)
Environment:
  JULIA_NUM_THREADS = 4
```
and CodecZlib v0.6.0, PackageCompiler v1.0.2.

KristofferC:
This is a Julia bug and will be fixed in Julia 1.4 which is coming out soon.

jaakkor2:
I just verified that I get the error with
```
julia> VERSION
v"1.4.0-rc2.0"
```
thus the fix ( https://github.com/JuliaLang/julia/pull/34754 ? ) is not completely effective.

KristofferC:
Thanks, seems https://github.com/JuliaLang/PackageCompiler.jl/issues/365 covers this

PR #366, opened 2020-03-05
add codeczlib to test

KristofferC:
This will likely not pass on Windows due to #360


KristofferC:
Oh yeah, this is not relocatable so it will fail for other reasons

PR #367, opened 2020-03-07
create new_sysimage_source_path in temp folder

seregaxvm:
When using Julia from Debian repository, `base_dir` is a write protected folder `/usr/share/julia/base/`. So, when creating a new app with `create_app()`, it fails due to lack of privileges. This PR moves temporary `new_sysimage_source_path` to the `tmp` directory, as it probably was intended.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -147,7 +147,7 @@ function create_fresh_base_sysimage(stdlibs::Vector{String}; cpu_target::String)
 
         # Use that to create sys.ji
         new_sysimage_content = rewrite_sysimg_jl_only_needed_stdlibs(stdlibs)
-        new_sysimage_source_path = joinpath(base_dir, "sysimage_packagecompiler_$(uuid1()).jl")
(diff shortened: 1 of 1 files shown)


KristofferC:
This is not correct, the current setup requires it to be in the same folder as the rest of the `Base` files.

seregaxvm:
Well, it worked on my machine. The only place `new_sysimage_source_path` is used is in [this call](https://github.com/JuliaLang/PackageCompiler.jl/blob/master/src/PackageCompiler.jl#L153-L155), where it's passed explicitly. After that it is [deleted](https://github.com/JuliaLang/PackageCompiler.jl/blob/master/src/PackageCompiler.jl#L159).

seregaxvm:
Just tested with `sysimage.jl` generated by `create_app`. It's not necessary to have `sysimage.jl` in the base folder as long as it's passed explicitly and current folder is `base_dir`. All these conditions are met.

KristofferC:
Oh that is great then! Thanks!

PR #380, opened 2020-04-08
Use dirname in stead of ../.. on the executable name for depot path.

rbvermaa:
Using the relative path (../..) concatenated to the executable
is not a valid path. We noticed this when the compiled
application opened an external julia process, which
tried to run a mkpath on the DEPOT_PATH, which was set
(via environment in src/embedded.c) to the concatenated
path, resulting in a obscure error:

  ERROR: IOError: mkdir: file already exists (EEXIST)

### DIFF
--- a/src/embedding_wrapper.c
+++ b/src/embedding_wrapper.c
@@ -9,6 +9,12 @@
 #include "uv.h"
 #include "julia.h"
 
+#ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


NHDaly:
Oh, nice find! Thanks @rbvermaa that makes sense and looks like it would fix the problem i saw, yeah. Thanks!

KristofferC:
Thanks

PR #401, opened 2020-05-06
document base_sysimage parameter to create_sysimage

grahamgill:
I'd like to have the `base_sysimage` parameter to `create_sysimage` documented as part of the API.

JuliaPy/pyjulia#380 proposes using `base_sysimage` to allow pyjulia to build its custom sysimage on top of an existing custom sysimage (e.g. produced using PackageCompiler). As long as `base_sysimage` isn't documented in the PackageCompiler API it isn't safe to rely on.

The only change proposed by the PR is to the `create_sysimage` doc string to document `base_sysimage`.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -355,6 +355,10 @@ by setting the environment variable `JULIA_CC` to a path to a compiler
 
 ### Advanced keyword arguments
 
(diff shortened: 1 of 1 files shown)


KristofferC:
https://github.com/JuliaPy/pyjulia/pull/380#issuecomment-931245732

PR #402, opened 2020-05-08
Wrong indention

newptcai:


fredrikekre:
It is correct, with this change the code block will move outside of the note.

Issue #413, opened 2020-06-11
Allow setting julia startup options on executable


NHDaly:
Either permanently during the build, or user-configurable at runtime, we would like to be able to set julia configuration options on the binary.

Specifically, we currently need the ability to set `-O1` on our build, since we're still experiencing an LLVM performance bug where it hangs at runtime when compiling new combinations of some input arguments. The details aren't terribly important, but it's basically a manifestation of this issue: https://github.com/JuliaLang/julia/issues/31419, and setting `-O1` skips the slow LLVM pass.

Can we either have the ability to set such arguments permanently during the build so that they'll always be set at runtime in the executable, and/or can we change the `embedding_wrapper.c` program to allow passing arguments to initialize julia with, just like Julia's standard c driver does?

Or yet a third option would be to allow specifying a custom C driver program, like PackageCompiler used to do before PackageCompilerX? Really any of those options would be fine for us in this case. Thanks @KristofferC.

--------

Also as more motivation, I could imagine for some applications, users may want to set `--compile=min` on the binary as well, since you'll have done almost all the compilation you actually want during your snooping, and any other random things that come up, like say plotting a new type of plot or something, you'd likely want to run without compiling?

I'm not sure if there are others, but i could imagine us still wanting to run the binary with e.g. `--check-bounds=yes`, `--track-allocation=all`, or different `--math-mode`s?

NHDaly:
Copying over @KristofferC's comment from https://github.com/JuliaLang/PackageCompiler.jl/pull/415#issuecomment-649608244:

> At some point, we should perhaps make the app automatically accept julia arguments like
> 
> ```
> ./app --julia-args -O1
> ```
> 
> etc.

---

+1

PR #415, opened 2020-06-18
Allow passing a custom c program to `create_app()`

NHDaly:
This will support users who need to set arbitrary julia jl_options for
julia_init. For example, configuring the binary to always run with `-O1`
or `--compile=min`.

It will of course then also support any other arbitrary changes to the C
program a user may want to make.

This is a partial fix for https://github.com/JuliaLang/PackageCompiler.jl/issues/413; it will allow us to write code to control the intializiation arguments ourselves. But as a follow-up, it might be nice to have a more complete fix where we add an option (maybe by select an alternate provided C program?) that will automatically parse command line args in the same was as the standard julia binary does.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -583,7 +583,7 @@ function create_app(package_dir::String,
                     filter_stdlibs=false,
                     audit=true,
                     force=false,
-                    c_driver_program::String="embedding_wrapper.c",
+                    c_driver_program::String=joinpath(@__DIR__, "embedding_wrapper.c"),
                     cpu_target::String=default_app_cpu_target())
     precompile_statements_file = abspath.(precompile_statements_file)
     precompile_execution_file = abspath.(precompile_execution_file)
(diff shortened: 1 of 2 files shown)


NHDaly:
I think the Travis failures were flakes, since only two tests failed, and both with Curl errors, a la `curl: (22) The requested URL returned error: 404 Not Found`

NHDaly:
I've fixed a bug, and this is now good to review.

Just noting that we are using this branch internally now at RelationalAI to force `-O1`, and it's working for us. :)
Thanks!

KristofferC:
At some point, we should perhaps make the app automatically accept julia arguments like

```
./app --julia-args -O1
```

etc.

NHDaly:
Thanks for the merge.

Yeah, agreed, regarding julia init args. That's why i've left open #413, since i think what you described would be a better fix. I'll leave a comment there pointing back to your comment here. 👍 Thanks!

PR #421, opened 2020-07-01
Use `import` to precompile packages (instead of using `using`)

DilumAluthge:
Both `import Foo` and `using Foo` will result in the precompilation of `Foo`.

`using`.:
- Pros: none
- Cons: awkward name clashes if one package exports a name that clashes with the name of another package (see e.g, `Future` and `Distributed.Future`)

`import`:
- Pros: no name clashes
- Cons: none


KristofferC:
Makes sense to me.

DilumAluthge:
Can we make a new release?

Issue #434, opened 2020-08-14
OhMyREPL sysimage breaks VScode language server


Crown421:
I am running julia 1.5-rc on Manjaro Linux and replaced my default sysimage with one including Revise and OhMyREPL. This was very nice, and Julia sessions started up very quickly. 

Unfortunately, with this sysimage the VScode Julia Language Server doesn't start, with the following error
```
Warning: Package OhMyREPL does not have REPL in its dependencies:
│ - If you have OhMyREPL checked out for development and have
│   added REPL as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with OhMyREPL
└ Loading REPL into OhMyREPL from project dependency, future warnings for OhMyREPL are suppressed.
fatal: error thrown and no exception handler available.
#<null>
require at ./loading.jl:906
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2214 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2398
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1690 [inlined]
call_require at /buildworker/worker/package_linux64/build/src/toplevel.c:425 [inlined]
eval_import_path at /buildworker/worker/package_linux64/build/src/toplevel.c:462
jl_toplevel_eval_flex at /buildworker/worker/package_linux64/build/src/toplevel.c:682
jl_parse_eval_all at /buildworker/worker/package_linux64/build/src/ast.c:913
jl_load_rewrite at /buildworker/worker/package_linux64/build/src/toplevel.c:914
include at ./Base.jl:380
include at ./Base.jl:368 [inlined]
include at /home/steffen/.julia/packages/OhMyREPL/7qVFi/src/OhMyREPL.jl:6 [inlined]
__init__ at /home/steffen/.julia/packages/OhMyREPL/7qVFi/src/OhMyREPL.jl:122
jfptr___init___32261 at /home/steffen/Apps/julia-1.5.0-rc2/lib/julia/sys.so (unknown line)
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2214 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2398
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1690 [inlined]
[...]

PR #439, opened 2020-09-04
Do not recommend replacing the default sysimage

mkitti:
Due to various issues with [compatibility with other packages](https://github.com/JuliaLang/PackageCompiler.jl/issues/434#issuecomment-675563737), do not recommend replacing the default sysimage. Propose alternatives such as an alias.

### DIFF
--- a/docs/src/sysimages.md
+++ b/docs/src/sysimages.md
@@ -97,10 +97,16 @@ Dict{Base.PkgId,Module} with 34 entries:
 ```
 
 Alternatively, instead of giving a path to where the new sysimage should appear, one
-can choose to replace the default sysimage.
-This is done by omitting the `sysimage_path` keyword and instead adding `replace_default=true`, for example:
+can choose to replace the default sysimage. **This is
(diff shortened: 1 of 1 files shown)


KristofferC:
I think we should even put a `@warn` in the code.

It would be good to provide a hint on the alternative which is to create an alias that aliases julia to something like `julia -J~/my/default/sysimage.so`.

docs/src/sysimages.md
```diff
@@ -119,6 +125,10 @@ REPL and package manager not require compilation after a Julia restart.. It is
 therefore unlikely that `incremental=false` is of much use unless in special
 cases for sysimage creation (for apps it is a different story though).
 
+As an alternative consider using another mechanism to pass the `-J` flag to
+Julia as above. These include creating a desktop shortcut or a shell alias,
+`$ alias julia='julia -q -J/path/to/sysimage.so'`, that includes the option.
+
```
mkitti, line 131: @KristofferC I recommend creating an alias in the last paragraph of the section.

PR #443, opened 2020-09-16
Ensure `@artifact` is skipped on non-windows platforms

### LINKED ISSUE #442: Artifacts stdlib broke artifact"..." macro ?
```
julia> using PackageCompiler
[ Info: Precompiling PackageCompiler [9b87118b-4619-50d2-8e1e-99f35a4d4d9d]
ERROR: LoadError: LoadError: Cannot locate artifact 'x86_64-w64-mingw32' in '/tmp/tmp.uZ4xRCy7sz/packages/PackageCompiler/vsMJE/Artifacts.toml'
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] artifact_slash_lookup(name::String, artifact_dict::Dict{String, Any}, artifacts_toml::String)
   @ Artifacts ~/julia-master/usr/share/julia/stdlib/v1.6/Artifacts/src/Artifacts.jl:543
 [3] @artifact_str(__source__::LineNumberNode, __module__::Module, name::Any)
   @ Artifacts ~/julia-master/usr/share/julia/stdlib/v1.6/Artifacts/src/Artifacts.jl:598
 [4] include
   @ ./Base.jl:379 [inlined]
 [5] include_package_for_output(input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, uuid_tuple::Tuple{UInt64, UInt64}, source::Nothing)
   @ Base ./loading.jl:1131
 [6] top-level scope
   @ none:1
 [7] eval
   @ ./boot.jl:345 [inlined]
 [8] eval(x::Expr)
   @ Base.MainInclude ./client.jl:446
 [9] top-level scope
   @ none:1
in expression starting at /tmp/tmp.uZ4xRCy7sz/packages/PackageCompiler/vsMJE/src/PackageCompiler.jl:103
in expression starting at /tmp/tmp.uZ4xRCy7sz/packages/PackageCompiler/vsMJE/src/PackageCompiler.jl:1
```

Looks like this line runs even though I am not on Windows:

https://github.com/JuliaLang/PackageCompiler.jl/blob/e560504d4dc7feaaed3e0585c6d0a7b79d5d3fc7/src/PackageCompiler.jl#L103

Pkg commit: https://github.com/JuliaLang/Pkg.jl/commit/c2c99807b157d6b9e23d09fa597800ec0e715cee
Julia commit: https://github.com/JuliaLang/julia/commit/8bdf5695e25e77d53f8934d9f135858d174e30d1

cc @staticfloat

staticfloat:
Because newer `@artifact` macros do more work at compile-time, we need to use a compile-time `if` statement here to avoid running the macro at all on non-windows systems.

Fixes #442

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -99,7 +99,7 @@ function get_compiler()
     if cc !== nothing
         return cc
     end
-    if Sys.iswindows()
+    @static if Sys.iswindows()
         return joinpath(Pkg.Artifacts.artifact"x86_64-w64-mingw32", "mingw64", "bin", "gcc.exe")
     end
     if Sys.which("gcc") !== nothing


fredrikekre:
Thanks. I suspected this was the problem. What work does it do during macro expansion, and why?

staticfloat:
It loads, parses and extracts the needed hash all at compile time, so as to save on time and space when loading the `.ji` file in the future.  Here's the commit, with the associated message: https://github.com/JuliaLang/julia/commit/2b2ee1f0afdadf49392d26f1035dee454617ec8f#diff-ccd5c99b75897c5269d89a687f8b0dc6

It only does that work at compile-time when the artifact name is a constant string; if you give it the result of a function call, or a variable or something like that, it will have to do the work at runtime.

tkf:
If you need a fix like this PR, doesn't it mean https://github.com/JuliaLang/julia/pull/37320 is backward incompatible?  Maybe `Pkg.Artifacts` should export the old run-time -only version of `@artifact_str`?  Also, I wonder if the compile-time constancy has a negative implication for relocatability: https://github.com/JuliaLang/julia/issues/37681

staticfloat:
I think this consequence is a small enough backwards compatibility that we don’t need to worry about it too much. It’s an easy enough fix.

tkf:
I understand "minor change" is a thing in Julia. But I'd like to note that https://semver.org does not says it's OK to introduce breaking change if it is easy to fix. Furthermore, it is trivial to make `Pkg.Artifacts.@artifact_str` backward-compatible because it is a different namespace than `Artifacts`.

(Technically, I guess it is possible to argue that this is not breaking since [`Pkg.Artifacts.@artifact_str` API documentation](https://julialang.github.io/Pkg.jl/v1.5/api/#Pkg.Artifacts.@artifact_str) does not specify when the key lookup happens. But given that most of Base/stdlib APIs are under-specified, I don't think it's OK to use this as an argument.)

Anyway, it's probably not a good idea for me to rant about SemVer compliance of `julia` here. You might have considered all the engineering trade-offs before breaking compatibility. I hope it's not going to introduce confusion in public and private projects using artifacts.

DilumAluthge:
Could we get a new release of PackageCompiler that includes this fix?

KristofferC:
Yep, done.

PR #454, opened 2020-10-07
Streaming read of precompile statement files

omus:
I noticed that when creating the system image we were reading all of the statements lines into memory. The only reason for this seems to be the `sort` call (which should be `sort!`) which shouldn't be necessary.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -197,17 +197,14 @@ function create_sysimg_object_file(object_file::String, packages::Vector{String}
                             isapp::Bool)
 
     # Handle precompilation
-    precompile_statements = ""
+    precompile_files = String[]
     @debug "running precompilation execution script..."
     tracefiles = String[]
     for file in (isempty(precompile_execution_file) ? (nothing,) : precompile_execution_file)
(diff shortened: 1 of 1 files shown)


KristofferC:
Out of curiosity, did this matter in practice for some workload?

omus:
> Out of curiosity, did this matter in practice for some workload?

I ran into a `LLVM ERROR: out of memory` when attempting to use a precompile statement file with ~15000 lines. The failure also split out this block of code which is where I noticed that the file contents were entirely being loaded into memory. I'm still working through my issue and I don't expect this change to help but this looked like a nice improvement.

omus:
Re-triggering CI

omus:
> I ran into a `LLVM ERROR: out of memory` when attempting to use a precompile statement file with ~15000 lines. The failure also split out this block of code which is where I noticed that the file contents were entirely being loaded into memory. I'm still working through my issue and I don't expect this change to help but this looked like a nice improvement.

This change didn't help my issue and I needed to increase the available memory on the instance in order to be able to generate a system image. Unfortunately the resulting system image when used results in a core dump.

As I mentioned I'd still like to merge this as a minor improvement.

PR #457, opened 2020-10-16
Log precompile _failures_ as well as errors. Pass log level through to helper process.

NHDaly:
This allows users to set the debug log level in the process, not just
via the `JULIA_DEBUG=all` environment variable (we _thought_ we had
enabled the debug messages, but we had not! 😬 ).

Also, adds a log statement for precompile statements that _fail_, not
just those that error. For example, this surfaces this failure:

```julia
┌ Debug: Precompilation failed: precompile(Tuple{typeof(Base.deepcopy_internal), Any, Base.IdDict{Any, Any}})
└ @ Main.anonymous none:31
```

Which is a pattern I've noticed in our own precompilation failures,
where Julia seems to be emitting a precompile statement with an abstract
argument (`Any` in this case, but i've seen others) for a method that is
_not_ marked `@nospecialize`, which causes `precompile()` to fail, since
you cannot normally instantiate a method for abstract types.

So that's a separate issue that we need to dig into, but this PR allows
us to identify issues like these. :)


NHDaly:
I would prefer to be able to somehow pass through the Log config if you set `JULIA_DEBUG=PackageCompiler`, but i couldn't figure out how to get that to work on the spawned process, because of the anonymous modules it creates.

I'd love suggestions for that if you have any! :)

src/PackageCompiler.jl
```diff
@@ -212,25 +214,40 @@ function create_sysimg_object_file(object_file::String, packages::Vector{String}
...
+                            @debug "Precompilation failed: \$statement"
+                        end
+                    catch e
+                        # See julia issue #28808
+                        errors += 1
+                        @debug "Error executing \$statement:\n\$e"
+                    end
                 end
+                total = length(precompile_statements)
+                @info "PackageCompiler: completed \$total precompile statements with \$failures failures, \$errors errors."
```
KristofferC, line 250: One problem here is that people think there is something wrong with their code when they see "x errors" but there is really nothing they can do to fix it. So I'm a little bit against adding this as something people see by default.

NHDaly: Yeah, i had the same concern, and wasn't sure what to do about it.

Maybe a flag, to control reporting precompilation success rate?

Or we could print this to `@debug`, too, but debug is very noisy and i think we'd like to see this all the time.

NHDaly: Do you have any suggestions? :) Thanks!

NHDaly: Maybe we can also print the link to the julia issue that's in the comment, above?

Or do you think I should just move this to a `@debug` log as well?

PR #459, opened 2020-10-19
Fix insufficient escaping

GunnarFarneback:
Add another backslash for sufficient escaping. The single backslash is fine in a regex but the regex is embedded within a regular string.


KristofferC:
Oops, thanks.

NHDaly:
Ah, damn, you beat me to it! 😁 

This was exposed by the logging added here: https://github.com/JuliaLang/PackageCompiler.jl/pull/457

Great fix! (If I might ask: how did you discover this? Do you have some tricks for manually or automatically reviewing the precompile statements generated by PackageCompiler)?

GunnarFarneback:
I have an embedded application which still has a startup time of 25 seconds with a system image so as a part of looking for low hanging fruit I hacked the logging of "failed to execute" to a `println` and noticed that the output looked funny (it was in Windows and there `"\1"` prints as something roughly similar to ⦿).

NHDaly:
Amazing. thanks for the detailed 👁️ (or ⦿? 😉)

PR #460, opened 2020-10-20
also call init_active_project when required

KristofferC:
Ref https://github.com/JuliaLang/julia/issues/37441


KristofferC:
Improves the situation on master but there is still something going on with the artifact stuff.

PR #463, opened 2020-10-23
 create_app: load "deps" packages besides just the app package

NHDaly:
I now understand the cause of `Case 5` in https://github.com/JuliaLang/julia/issues/28808#issuecomment-712472418, and this PR is a first attempt at addressing it.

Before this commit, `create_app()` would only load the package for the
App being compiled before executing all precompile statements. This
means that if your precompile script was importing any "extra" packages,
that aren't used directly by the app's module, that they would silently
fail to statically compile.

We are doing this in our static compilation job to also precompile some "helper" packages that we ship alongside our application code, such as `Plots`.

I'm not sure what the best way to approach this fix is. Some options I see:
- Add a parameter to `create_app` to allow optionally specifying extra packages to include in the static compilation
- Automatically also include _all packages_ from the app's Project.toml (that's the approach I took in this PR, so far, but happy to change it)
- Add a mechanism to the execution of the snoop precompilation scripts to report out the `Base.loaded_modules` at the end of the execution, and add those to the packages.
    - I like this option the best, because things Just Work how the user would expect them to: "If I add this to my snoop file, it will get statically compiled," but I think it's slightly complicated. Probably not too hard though! :)

I would love your feedback! Thanks @KristofferC!

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -4,6 +4,7 @@ using Base: active_project
 using Libdl: Libdl
 using Pkg: Pkg
 using UUIDs: UUID, uuid1
+using Logging
 
 export create_sysimage, create_app, audit_app, restore_default_sysimage
 
@@ -212,6 +213,12 @@ function create_sysimg_object_file(object_file::String, packages::Vector{String}
     precompile_code = """
         # This @eval prevents symbols from being put into Main
         @eval Module() begin
+
+            using Logging
+            # Pass through the user's logging level to the spawned process.
+            logger = ConsoleLogger(stderr, Logging.$(Logging.min_enabled_level(global_logger())))
+            global_logger(logger)
+
             PrecompileStagingArea = Module()
             for (_pkgid, _mod) in Base.loaded_modules
                 if !(_pkgid.name in ("Main", "Core", "Base"))
@@ -226,10 +233,13 @@ function create_sysimg_object_file(object_file::String, packages::Vector{String}
                 # N with a large number seems to work around it.
                 statement = replace(statement, r"Vararg{(.*?), N} where N" => s"Vararg{\1, 100}")
                 try
-                    Base.include_string(PrecompileStagingArea, statement)
-                catch
+                    success = Base.include_string(PrecompileStagingArea, statement)
+                    if !success
+                        @debug "Precompilation failed: \$statement"
+                    end
+                catch e
                     # See julia issue #28808
-                    @debug "failed to execute \$statement"
+                    @debug "Error executing \$statement:\n\$e"
                 end
             end
         end # module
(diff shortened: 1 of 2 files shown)


KristofferC:
I see, yes, deterministically loading all deps makes sense to me for `create_app`.

src/PackageCompiler.jl
```diff
@@ -619,6 +640,14 @@ function create_app(package_dir::String,
     bundle_julia_libraries(app_dir)
     bundle_artifacts(ctx, app_dir)
 
+    # Load all packages in the Project into the sysimg for Apps. Without specifying this,
+    # we cannot serialize any methods from those packages, because the packages won't be
+    # loaded in the final julia process. Since we cannot know which packages will be needed
+    # we can load _all_ of the packages (and their recursive packages) from the project.
+    # This may be more than the user will actually need, but that is up to users to prune
+    # their Project files to only those `deps` actually in use.
+    packages = Symbol.(keys(Pkg.TOML.parsefile(ctx.env.project_file)["deps"]))
```
KristofferC, line 649:
```
julia> ctx.env.project.deps
Dict{String,Base.UUID} with 37 entries:
  "BinaryBuilder"           => UUID("12aac903-9f7c-5d81-afc2-d9565ea332ae")
  "Documenter"              => UUID("e30172f5-a6a5-5a46-863b-614d45cd2de4")
  "ForwardDiff"             => UUID("f6369f11-7733-5829-9624-2563aa707210")
...
```

is probably better

src/PackageCompiler.jl
```diff
@@ -265,7 +286,7 @@ function create_sysimg_object_file(object_file::String, packages::Vector{String}
 
     if isapp
         # If it is an app, there is only one packages
-        @assert length(packages) == 1
+        #@assert length(packages) == 1
```
KristofferC, line 289: Just remove

NHDaly:
```suggestion
```

src/PackageCompiler.jl
NHDaly:
```suggestion
    packages = keys(ctx.env.project.deps )
```
Awesome thanks. Does this look good? ⬆️

NHDaly (commented):
Thanks for the review. Cool, if this works for you, then it works for me! :)

src/PackageCompiler.jl
NHDaly: Oops, I guess that should be an array of symbols:
```julia
    packages = Symbol.(keys(ctx.env.project.deps))
```
Fixed in https://github.com/JuliaLang/PackageCompiler.jl/pull/463/commits/af09bd8b7db975d818366998d889439b20141e2f

NHDaly:
Also, this PR is currently based on https://github.com/JuliaLang/PackageCompiler.jl/pull/457. I guess we should review that first before merging this one. :) Thanks for your help, @KristofferC! 😁

KristofferC:
I've also noticed this problem when creating a sysimage with Plots. If GR is set to use GR_jll it only loads GR_jll when actually executing the first plot. That means that GR_jll & dependencies will not be included in the sysimage since the sysimage only does a `using Plots`.

PR #471, opened 2020-12-05
Update for 1.6 linking and handle stdlibs that aren't in the sysimage

ianshmean:
Doesn't work yet. Attempts to handle stdlibs that aren't in the sysimage 

Trying to fix #470 

@KristofferC I thought it might be helpful to start this. Do you see where the remaining work is needed?

PR #472, opened 2020-12-06
Add github actions tests

ianshmean:
I added github actions tests in #471 but probably best to check master first


ianshmean:
The last CI run on master was with Travis on 1.5.2, which was successful.
GHA shows parity in success on 1.5.2 and shows regression in tests in 1.5.3

Issue #464, opened 2020-10-23
Allow skipping download of lazy artifacts


rbvermaa:
When using PackageCompiler.jl, we ran into an issue with TimeZone.jl [0] where it downloaded all timezone data (200+ files).  These files are listed as `lazy`. We would like to avoid downloading them, and in the issue linked, it was mentioned we should probably add an `include_lazy_artifacts` option.

I tried to add an `include_lazy_artifacts` option [1], however, it is probably not a correct solution, as it ignores the situation where there is a `Dict` in the Artifact section. See first part of the if at https://github.com/RelationalAI-oss/PackageCompiler.jl/blob/rv-include-lazy/src/PackageCompiler.jl#L722 . 

This patch, although probably incorrect, solved our immediate issue (we ger blocked by the server when downloading too many files).

I think this probably only can be fixed in Pkg.jl, but not sure how. Any idea how to get this fixed properly?

[0] https://github.com/JuliaTime/TimeZones.jl/issues/300
[1] https://github.com/JuliaLang/PackageCompiler.jl/compare/master...RelationalAI-oss:rv-include-lazy

NHDaly:
CC: @staticfloat  - i think we talked with you and @KristofferC about this on slack. Do you think something like an `include_lazy_artifacts` option is the best approach? Should it be along the lines of @rbvermaa's patch, above?

Thanks!

KristofferC:
It makes sense to me to have that option, at least.

ianshmean:
This is quite painful for CUDA.jl given all of the many large artifacts are lazy https://github.com/JuliaGPU/CUDA.jl/blob/master/Artifacts.toml

And a `include_lazy_artifacts` flag wouldn't help here, I believe. I think PackageCompiler will need to know which lazy artifacts it needs to grab in advance

Sacha0:
Bumping as this issue bit us again just now. Given

> It makes sense to me to have that option, at least.

I'll roll Rob's branch into a pull request so that we can begin iterating something. Thanks folks! :)

KristofferC:
This is now merged into PackageCompiler

PR #489, opened 2021-01-07
Add include_lazy_artifacts flag to create_app.

Sacha0:
This pull request reconstitutes @rbvermaa's stab at https://github.com/JuliaLang/PackageCompiler.jl/issues/464 in https://github.com/RelationalAI-oss/PackageCompiler.jl/tree/rv-include-lazy. Rob notes

> I tried to add an include_lazy_artifacts option [1], however, it is probably not a correct solution, as it ignores the situation where there is a Dict in the Artifact section. See first part of the if at https://github.com/RelationalAI-oss/PackageCompiler.jl/blob/rv-include-lazy/src/PackageCompiler.jl#L722 .

which I haven't looked into yet. But putting this pull request up to get the ball rolling :).

Co-Authored-By: Rob Vermaas <rob.vermaas@gmail.com>

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -618,7 +618,8 @@ function create_app(package_dir::String,
                     audit=true,
                     force=false,
                     c_driver_program::String=joinpath(@__DIR__, "embedding_wrapper.c"),
-                    cpu_target::String=default_app_cpu_target())
+                    cpu_target::String=default_app_cpu_target(),
+                    include_lazy_artifacts=true)
     precompile_statements_file = abspath.(precompile_statements_file)
     precompile_execution_file = abspath.(precompile_execution_file)
     package_dir = abspath(package_dir)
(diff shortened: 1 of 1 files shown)


Sacha0:
Complete solution or no, CI seems happy! 😄

Sacha0:
Friendly bump. Thoughts? :)

fguevaravas:
I think this is a great idea and will test this PR in my usage case and report back in a few days. 

I have issues with apps generated with PackageCompiler.jl getting huge because every single artifact that might be remotely needed is downloaded. To give an example: compiling an app that uses FFTW.jl, downloads the whole Intel MKL and OpenMP even if the app never ends up using these (we're talking about 1.5GB of things that don't need to be packaged with app). Ditto for all the timezone data in TimeZone.jl

I am concerned about one thing though, what is the behavior if there is an artifact that is really needed by the app and it wasn't downloaded at compile time? Will the app download it and crash if it can't? If that is the case, it may be better to have a more granular control? Perhaps providing a list of artifacts to include/exclude?

jlperla:
Sorry to bump this, but is this PR safe to use (or is there a good reason it hasn't been merged)?  We have been bitten by the https://github.com/JuliaTime/TimeZones.jl/issues/325  issue.

Sacha0:
We have been using this patch continuously for several months now, for what that's worth :). IIRC this pull request is waiting on formation of consensus on how best to move forward (with this or another approach to addressing the underlying issue).

KristofferC:
Sorry, yes, we should clearly get this in.

KristofferC:
https://github.com/JuliaLang/PackageCompiler.jl/pull/529 should hopefully do it.

Issue #392, opened 2020-04-22
creating *.dll from Julia code


OndWolf:
Dear contributors,
Package version 0.6.4 have function build_shared_lib(...). The goal for me is to create some code in Julia (I really like the syntax) and then have a possibility of calling this in C++ for example. One of wish for future was to try it in Matlab/Simulink (call from s-function). Unfortunately build_shared_lib(...) function is not include in last version 1.1.1. Is there chance that this could be in PackageCompiler in future again? Or maybe there is another possibility of having Julia code available in another enviroment?

Here is example of how I use it:
Julia code:

```
Base.@ccallable function test()::Cint
    return(2)
end
```

then create dll as:

build_shared_lib("C:/Users/Documents/Julia/test.jl", "test")

and call it as:
```
#include <Windows.h>

typedef int(*fcnptr)();
fcnptr myFnc;

int main()
{
	HMODULE handle = LoadLibrary(L"test.dll");
	if (handle != NULL)
	{
		myFnc = (fcnptr)GetProcAddress(handle, "test");
		std::cout << myFnc() << std::endl;
	}
	else
	{	
		std::cout << "Failed!\n";
	}
}
```

Thank you for answer.

LeonardJohard:
I have the same request for interoperability with Modelica. In principle a dynamic library solution is much more useful than compiling into executables in our use cases. Any rough estimations or roadmaps for this functionality?

KristofferC:
> Package version 0.6.4 have function build_shared_lib(...). The goal for me is to create some code in Julia (I really like the syntax) and then have a possibility of calling this in C++ for example. One of wish for future was to try it in Matlab/Simulink (call from s-function). Unfortunately build_shared_lib(...) function is not include in last version 1.1.1.

A sysimage is already a shared library. So try something like

```
build_sysimage(; sysimage_path="test.so", script="test.jl")
```

(the `script` arguments inserts that code into the sysimage).

You can see that the symbol is in there:

```
~/julia kc/devdocs*
❯ nm sys.so | grep test
...
0000000000135ca0 T test
```

Note that you will also in general need to initialize the Julia runtime.

simonbyrne:
Here is an example of using PackageCompiler to build a shared library:
https://github.com/simonbyrne/cg/

KristofferC:
https://github.com/JuliaLang/PackageCompiler.jl/pull/490

KristofferC:
[...]

Issue #350, opened 2020-02-21
Dynamic Linking for shared library in PackageCompilerX


prakhar2b:
In the new PackageCompilerX, we build a relocatable executable by copying julia library files in the same folder as executable and dynamically link it with that using `rpath` and  `$ORIGIN`

```Bash
gcc -DJULIAC_PROGRAM_LIBNAME=\"sys.so\" -o MyApp MyApp.c sys.so -O2 -I'/home/kc/julia/include/julia' -L'/home/kc/julia/lib' -fpie -Wl,-rpath,'$ORIGIN:$ORIGIN/../lib' -ljulia
```

However, the shared library `sys.so` that we created is still linked with local julia installation, and can't be loaded (using `ctypes.CDLL` in Python).

We can solve that by using same workaround for sysimage linking that we did for executable.

```Bash
gcc -shared -o sys.so -Wl,--whole-archive sys.o -Wl,--no-whole-archive -L"/home/kc/julia/lib" -fpie -Wl,-rpath,'$ORIGIN:$ORIGIN/../lib'  -ljulia
```

However, we still can't use it until we initialize the julia runtime  before calling any ccallable function. In earlier version of PackageCompiler, we would have `lib_init.c`for that. Should we not have something like that here as well?

KristofferC:
Yes, we should probably add an argument to `create_sysimage` that adds these init and exit runtime functions to it.

KristofferC:
I tried with the old packagecompiler with `ctypes.CDLL` and it didn't really work:

```
❯ python 

Python 2.7.16 (default, Dec 13 2019, 18:00:32) 
[GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.32.4) (-macos10.15-objc-s on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from ctypes import *

>>> lib = CDLL("Users/kc/PackageCompiler.jl/builddir/foo.dylib")

>>> lib.init_jl_runtime()
ERROR: could not load library "/System/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/foo.dylib"
dlopen(/System/Library/Frameworks/Python.framework/Versions/2.7/Resources/Python.app/Contents/MacOS/foo.dylib, 6): image not found
```

Does it need a full path?

prakhar2b:
```Python
[...]

PR #490, opened 2021-01-09
Add `create_library` function

### LINKED ISSUE #257: Question: How do we use build_shared_lib function Properly ?
I'm trying to generate shared library using `build_shared_lib` which  PackageCompiler.jl provides.

Here is my sample code named `example.jl`

```julia
module Example

Base.@ccallable empty()::Cint = 10
Base.@ccallable foo(x::Cint)::Cint = 3
Base.@ccallable function bar(x::Cint)::Cint
    return 3
end

Base.@ccallable function just_print()::Cvoid
    print("Example")
end

Base.@ccallable function bar_segfault(x::Cint)::Cint
    println("Example")
    return 3
end

Base.@ccallable function example(x::Cint)::Cint
    return x
end
end
```

Apparently building shared object succeed by using the following the script named `gen_lib.jl`:

```julia
using PackageCompiler
build_shared_lib("example.jl")
```

```console
$ julia gen_lib.jl
```

This will create `builddir/example.dylib`.

The problem is some function works, but other cause segmentation fault.

Here is a code to reproduce our problem:

- environment

```julia
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.1.1 (2019-05-16)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> versioninfo()
Julia Version 1.1.1
Commit 55e36cc (2019-05-16 04:10 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin15.6.0)
  CPU: Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
```

- make call script named `call_example.jl`

```julia
const dylibpath="builddir/example"
@show ccall((:foo,dylibpath),Cint,(Cint,),100)
@show ccall((:bar,dylibpath),Cint,(Cint,),100)

println("next :bar_segfauld will be seg fault...")

@show ccall((:bar_segfault,dylibpath),Cint,(Cint,),100)
```

```console
$ julia call_example.jl
ccall((:foo, dylibpath), Cint, (Cint,), 100) = 3
ccall((:bar, dylibpath), Cint, (Cint,), 100) = 3
next :bar_segfauld will be seg fault...

signal (11): Segmentation fault: 11
in expression starting at /Users/terasaki/tmp/sample/call_example.jl:7
unknown function (ip: 0xffffffffffffffff)
Allocations: 295934 (Pool: 295852; Big: 82); GC: 0
Segmentation fault: 11
```

How do we use shared object Properly ?

kmsquire:
* Similar to `create_app`, this function facilitates the creation of
  C shared library bundles using Julia.
* As with apps, it generates a directory tree with required shared
  libraries and artifacts.
* In contrast with apps, the julia system image is the main artifact
  generated, and its name should be set to the expected name of
  the library you're creating.
* One downside of this implementation is that any binaries linking
  to this shared library may need help finding it at runtime.  The
  methods for doing so are system dependent.

Probably fixes #257, #350, #392.

Still needs documentation (at least).  There is a bit of explanation in the function docs for `create_library` that should probably be moved elsewhere.

Companion MR in https://github.com/simonbyrne/libcg/pull/18

Cc: @simonbyrne


src/PackageCompiler.jl
```diff
@@ -619,6 +639,130 @@ function create_app(package_dir::String,
...
+                    precompile_statements_file,
+                    incremental,
+                    filter_stdlibs,
+                    audit,
+                    force,
+                    cpu_target::String;
+                    library_only::Bool=false,
+                    c_driver_program::String="",
+                    julia_init_c_file::String="",
+                    header_files::Vector{String}=[],
```
simonbyrne, line 762:
```suggestion
                    header_files::Vector{String}=String[],
```

src/julia_init.c
```diff
@@ -0,0 +1,74 @@
...
+#include <libgen.h>
+#endif
+
+// Julia headers (for initialization and gc commands)
+#include "uv.h"
+#include "julia.h"
+
+#include "julia_init.h"
+
+JULIA_DEFINE_FAST_TLS();
```
KristofferC, line 24: I don't think this should be defined for libraries (see the comment in the code at https://docs.julialang.org/en/v1/manual/embedding/#High-Level-Embedding).

simonbyrne: What is `JULIA_DEFINE_FAST_TLS`?

src/PackageCompiler.jl
```diff
@@ -786,4 +941,16 @@ function bundle_artifacts(ctx, app_dir)
     return
 end
 
+function bundle_headers(dest_dir, header_files)
+    isempty(header_files) && return
+    include_dir = joinpath(dest_dir, "include")
+    mkpath(include_dir)
+
+    for header_file in header_files
+        new_file = joinpath(include_dir, basename(header_file))
```
kmsquire, line 950: This assumes all header files are at the top level.  For basic use cases, this is probably fine, but it might be good to allow paths relative to the top level include.  Not sure of the best way to do this...

kmsquire:
Status: based on the PR to @simonbyrne's [libcg](https://github.com/simonbyrne/libcg/pull/18), this seems to be working on Linux and OSX fine in Julia v1.5.  For Windows, the PATH needs to be set, and I'm not quite sure of the right incantation in the `ci.yml` file yet.

Julia v1.6 can't be tested in CI yet because of [this bug](https://github.com/julia-actions/setup-julia/issues/66), but `latest` is a reasonable proxy.  On the latest master, libcg isn't yet linked to `libjulia-internal`--need to copy the relevant lines from `PackageCompiler.jl` and figure out how to condition on the Julia version in the libcg `Makefile` (any hints welcome!).

As above, the companion PR in libcg is simonbyrne/libcg#18.

src/julia_init.c
kmsquire: Removed in a recent commit.

src/PackageCompiler.jl
kmsquire: Fixed and pushed locally.

ianshmean:
@kmsquire FYI master now includes CI for the 1.6 beta1, if you wanted to rebase

docs/src/libs.md
```diff
@@ -0,0 +1,120 @@
+# [Libraries](@id libraries)
+
+Creating a library with PackageCompiler involves creating a custom system image with a couple of
+additional features to faciltate linking and use by external (non-Julian) programs--it's already
```
Byrth, line 4: Facilitate is missing an `i`

docs/src/libs.md
```diff
@@ -0,0 +1,120 @@
...
+libjulia.1.6.dylib
+libjulia.1.dylib
+libjulia.dylib
+...
+```
+
+(These will have an `.so` extension on Linux, and a `.dll` extension on Windows.  There may also
+be other files in the same directory, depending on your operating system and version of Julia.)
+
+In addtion to most of the same keyword arguments as
```
Byrth, line 77: addition is missing an `i`

src/PackageCompiler.jl
```diff
@@ -432,11 +509,13 @@ function create_sysimage(packages::Union{Symbol, Vector{Symbol}}=Symbol[];
         end
         base_sysimage = create_fresh_base_sysimage(sysimage_stdlibs; cpu_target=cpu_target)
     else
-        if base_sysimage == nothing
+        if base_sysimage === nothing
```
Byrth, line 512: Is using `isnothing()` more Julian?

Byrth (commented):
Just got interested in the PR and skimmed it. Nice work!

docs/src/libs.md
kmsquire: Fixed.

docs/src/libs.md
kmsquire: Fixed.

src/PackageCompiler.jl
kmsquire: I'm not sure, but I made this change.  Thanks for reviewing, @Byrth!

PR #503, opened 2021-03-26
remove `Base.@ccallable` from doc examples

visr:
After upgrading from julia 1.5 to 1.6 my app was no longer compiling, with the error:

```
ERROR: @ccallable was already defined for this method name
```

In the docs I then read that this annotation is no longer needed, however I still saw it in use in the docs. Hence I now remove the annotation from the docs.

For reference, a C callable function is still created, but it wraps the module's `julia_main`, see
https://github.com/JuliaLang/PackageCompiler.jl/blob/929657b22f3350ff1ed631ceb27dc6a715ef9a37/src/PackageCompiler.jl#L283-L293

### DIFF
--- a/docs/src/devdocs/binaries_part_2.md
+++ b/docs/src/devdocs/binaries_part_2.md
@@ -19,8 +19,8 @@ section (e.g. the `julia-config.jl` script) but it is good to know they exist.
 
 A rough outline of the steps we will take to create an executable are:
 
-- Create our Julia app with a `Base.@ccallable` entry-point which means the Julia
-  function can be called directly from C.
+- Create our Julia app with a `julia_main()::Cint` entry-point which can be
+  called directly from C.
 - Create a custom sysimage to reduce latency (this is pretty much just doing
   part 1) and to hold the C-callable function from the first step.
 - Write an embedding wrapper in C that loads our custom sysimage, does some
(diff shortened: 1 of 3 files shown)


KristofferC:
These are devdocs and are just meant to explain how PackageCompiler works. It has nothing really to do with the usage of PackageCompiler. That's why there is a

> This section is for people who want to understand PackageCompiler.jl under
    the hood. It is not required reading to use the package.

note on the top of these pages.

visr:
Yeah that's true, but it will probably still confuse people if the examples there do not work.

I realize there is a correct example here: https://github.com/JuliaLang/PackageCompiler.jl/tree/master/examples/MyApp

But as far as I'm aware it is not linked from the docs.
image

Since the examples in the docs only cover sysimage creation and not apps, I'm probably not the only one copying the example from the devdocs.

KristofferC:
Yeah, I can see that but I don't think the dev docs will work with this removed. The reason it is not needed in PackageCompiler is because of 

https://github.com/JuliaLang/PackageCompiler.jl/blob/929657b22f3350ff1ed631ceb27dc6a715ef9a37/src/PackageCompiler.jl#L284-L291

but this is not present in the devdocs. So someone following the devdocs will get an error with this change. An example with creating an app should probably be made and maybe the devdocs should be a bit more hidden so they don't look so tempting for normal usage :P

PR #512, opened 2021-05-26
Adapt to changes in Pkg.Operation method signatures under Julia 1.7.

Sacha0:
PackageCompiler calls Pkg.Operation.{load_all_deps,source_path} with a Pkg.Types.Context, valid under Julia 1.6. Under Julia 1.7, Pkg.Operation.load_all_deps requires a Pkg.Types.EnvCache, and Pkg.Operation.source_path requires a String. This patch adds version guards around these calls that hopefully pass the right objects. Best! :)


Sacha0:
Thanks for reviewing/merging Kristoffer! :)

PR #520, opened 2021-07-20
Fix: Incorrect tree hash in Artifacts.toml

vdayanand:


KristofferC:
I've verified on Windows that this is the correct hash.

PR #526, opened 2021-08-10
Propogate ARGS into precompile_execution_file

rssdev10:
The patch provides delivering of the Julia arguments into the children process which is used for precompiling execution files.

```bash
julia --project=@. --startup-file=no -e '
using PackageCompiler;
PackageCompiler.create_app(pwd(), "bin";
    precompile_execution_file=["src/run.jl"])
' create some_task
```

instead of:
```bash
julia --project=@. --startup-file=no --trace-compile=build/precompile_statements.jl src/run.jl create some_task

julia --project=@. --startup-file=no -e '
using PackageCompiler;
PackageCompiler.create_app(pwd(), "bin";
    precompile_statements_file=["build/precompile_statements.jl"])
'
```


KristofferC:
`ARGS` here is a bit odd imo. I would say this is subsumed by https://github.com/JuliaLang/PackageCompiler.jl/pull/569.

PR #529, opened 2021-09-10
Rebase: Add include_lazy_artifacts flag to create_app (issues/464).

KristofferC:
cc @Sacha0, does this look ok to you?

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -746,6 +746,9 @@ compiler.
 
 - `force::Bool`: Remove the folder `compiled_app` if it exists before creating the app.
 
+- `include_lazy_artifacts::Bool`: if lazy artifacts should be included in the bundled artifats,
+  defaults to `true`.
+
 ### Advanced keyword arguments
 
 - `cpu_target::String`: The value to use for `JULIA_CPU_TARGET` when building the system image.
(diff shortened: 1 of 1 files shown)


src/PackageCompiler.jl
```diff
@@ -760,7 +763,8 @@ function create_app(package_dir::String,
                     audit=true,
                     force=false,
                     c_driver_program::String=joinpath(@__DIR__, "embedding_wrapper.c"),
-                    cpu_target::String=default_app_cpu_target())
+                    cpu_target::String=default_app_cpu_target(),
+                    include_lazy_artifacts::Bool=true)
```
Sacha0, line 767: I imagine I'm missing something, but at first blush it seems that this argument should be passed to the `_create_app` call in this method's body? :)

src/PackageCompiler.jl
```diff
@@ -851,6 +855,9 @@ compiler.
    Used to determine and set the `compatibility_version` (on Apple) and `soname` (on
    Linux/UNIX).
 
+- `include_lazy_artifacts::Bool`: if lazy artifacts should be included in the bundled artifats,
```
Sacha0, line 858: `artifats` 😄

src/PackageCompiler.jl
```diff
@@ -746,6 +746,9 @@ compiler.
 
 - `force::Bool`: Remove the folder `compiled_app` if it exists before creating the app.
 
+- `include_lazy_artifacts::Bool`: if lazy artifacts should be included in the bundled artifats,
```
Sacha0, line 749: `artifats` here too 😄

Sacha0:
The only potentially important comment is the the first, in case I wasn't missing something there :).

src/PackageCompiler.jl
kmsquire: I don't think you're missing anything--it definitely should be passed.

MR: https://github.com/JuliaLang/PackageCompiler.jl/pull/530

PR #530, opened 2021-09-13
Add missing `include_lazy_artifacts` parameter

kmsquire:


Sacha0 (approved):
Thanks Kevin! :)

Not sure about the CI failures? (Haven't looked properly.)

KristofferC:
Thanks. What I usually do to avoid things like this is to not set any default keyword values on "inner" functions and enforce that all keyword values are passed from the outer, user-exposed functions. I find that otherwise it is way too easy to miss passing on a keyword like this. By leaving off the default keyword value you get an error instead.

kmsquire:
> What I usually do to avoid things like this is to not set any default keyword values on "inner" functions and enforce that all keyword values are passed from the outer, user-exposed functions.

I believe I created that "inner" function and left the default there, and I like your suggestion.  I'll make a PR to change it.

PR #476, opened 2020-12-11
Fix empty sysimage test on 1.5.3. Fix app test on nightly

ianshmean:
Tidied up summary:
- On 1.5.3 `temp_cleanup_purge` was failing to compile at exit during generation of a blank sysimage, so this clears the `_atexit` activites during the sysimage generation subprocess. This is more of a bandaid than a fix. I couldn't figure out why it was failing

- The checked-in Manifest for the test app was causing stdlib jll problems, so the Manifest has been deleted and an instantiate step has been added to `create_app`. On nightly ubuntu a deadlock was occurring when autoprecomp was enabled after instantiate, but not on macos. I couldn't figure out why, but I disabled autoprecomp

A windows failure remains during the first app test. I have no clues on that one (apart from it appearing to be another `_atexit`-related issue)

PR #531, opened 2021-09-14
Disable autoprecomp during project instantiation

kmsquire:
* This was previously done for one call to instantiate(), but not to a second
  call.  Without this, precompilation is failing for most packages
  in `create_app()` or `create_lib()`.

The previous change was in #476.  There, disabling precompilation fixed a deadlock on Ubuntu, but didn't seem to have a visible effect on OSX.  Here, it fixes precompilation on OSX (at least).  I'm also unsure why it's necessary.

Cc: @IanButterworth @KristofferC

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -503,7 +503,11 @@ function create_sysimage(packages::Union{Symbol, Vector{Symbol}}=Symbol[];
     # Instantiate the project
     ctx = create_pkg_context(project)
     @debug "instantiating project at $(repr(project))"
-    Pkg.instantiate(ctx, verbose=true)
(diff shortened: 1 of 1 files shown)


kmsquire:
@KristofferC If you remember, I actually mentioned this issue to you before JuliaCon.

KristofferC:
Thanks, yes, this makes sense to me.

kmsquire:
> I don't realize why auto precompile would fail though.

I'm really not sure why either.  It bears looking into, but not something I have time for, unfortunately.

IanButterworth:
At what point does it hang? Before starting? On the last package? etc

KristofferC:
I don't think it hangs. It just reports that all packages fail to precompile.

PR #533, opened 2021-09-14
Remove keyword arguments from inner _create_app function

kmsquire:
* This makes it less likely to miss passing in a required argument.


KristofferC:
They can still be keyword arguments, just enough to leave out the default value (which enforces that the keyword argument is provided).

kmsquire:
Ah, ok--I was wondering about that. I'll change then back to keyword
arguments.

kmsquire:
Changed back to keyword arguments without defaults.

Sacha0 (approved):
Looks great, thanks Kevin! :)

PR #536, opened 2021-09-15
Remove Manifest.toml from MyLib/build

kmsquire:
This ensures that we will always use the most recent version of
PackageCompiler, rather than the version in Manifest.toml.


KristofferC:
It could also have a relative devved path (`pkg> dev../..` in that folder).

PR #537, opened 2021-09-16
dev PackageCompiler in MyLib/build

KristofferC:
This ensures that we are using the PackageCompiler in of the repository.

### DIFF
--- a/examples/MyLib/build/Manifest.toml
+++ b/examples/MyLib/build/Manifest.toml
@@ -1,0 +1,116 @@
+# This file is machine-generated - editing it directly is not advised
+
+[[ArgTools]]
+uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
+
+[[Artifacts]]
+uuid = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
+
+[[Base64]]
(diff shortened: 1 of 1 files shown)


examples/MyLib/build/Manifest.toml
```diff
@@ -0,0 +1,116 @@
...
+
+[[MozillaCACerts_jll]]
+uuid = "14a3606d-f60d-562e-9121-12d972cd8159"
+
+[[NetworkOptions]]
+uuid = "ca575930-c2e3-43a9-ace4-1e988b2c1908"
+
+[[PackageCompiler]]
+deps = ["Libdl", "Pkg", "UUIDs"]
+path = "../../.."
```
KristofferC, line 62: note

kmsquire (approved):
Thanks!

kmsquire:
Ah, I had forgotten the other reason that I removed the `Manifest.toml` file: it's (possibly) incompatible between minor Julia versions--i.e., this manifest file probably won't work for Julia 1.5 (I've discovered that locally with other projects), and may not work for Julia 1.7.

So I guess the "to do" here is to test this on Julia 1.7, except that we're still broken there.

PR #538, opened 2021-09-16
put the sysimage for apps inside lib and slightly tweak rpath

KristofferC:
this allows the sysimage in `create_library` to not live in a folder called `lib` which is sometimes useful

Previously, if one renamed the `lib` folder that is created by `create_library`, the dynamic linker would not find e.g. `libjulia-internal` since it looked in `../lib/julia`. Now it just looks in `./julia`.


kmsquire (approved):
Nice!

PR #539, opened 2021-09-16
remove some old code now that we only support julia 1.6+

KristofferC:


Sacha0 (approved):
Looks great! :)

Sacha0:
CI appears happy (modulo nightly, as expected given the pending FAST_TLS change), so this patch appears mergeworthy :).

PR #540, opened 2021-09-16
update to new syntax for JULIA_DEFINE_FAST_TLS

KristofferC:
Note sure this is the best way to do it.


Sacha0 (approved):
Looks like it should do the trick :).

Sacha0:
The nightly x86 windows test failure looks unrelated, so this patch should be in shape to merge I think? :)

PR #541, opened 2021-09-16
hide the "devdocs" in the manual

KristofferC:
I am not sure these are very useful to have anymore. They are slowly becoming outdated and most people just seem confused about if they should use the stuff written in there over the normal PackageCompiler interface.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -19,12 +19,14 @@ makedocs(
             "examples/plots.md",
         ],
 
+        #=
         "PackageCompiler - the manual way" => [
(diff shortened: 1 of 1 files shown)


tfiers:
I found an [old hosted version](https://docs.juliahub.com/PackageCompiler/MMV8C/1.2.1/devdocs/intro.html) of these dev docs through a google query and found them very interesting.

KristofferC:
They are supposed to be useful yes but people always skipped reading the part where it says that they are not a tutorial for how to use the package and then complained that they couldn't get things to work.

PR #542, opened 2021-09-17
remove more pre 1.6 code and simplify kwarg passing

KristofferC:
Just a small cleanup that is possible now when we moved to Julia 1.6.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -121,7 +121,7 @@ end
 
 function get_julia_cmd()
     julia_path = joinpath(Sys.BINDIR, Base.julia_exename())
-    cmd = `$julia_path --color=yes --startup-file=no`
+    return `$julia_path --color=yes --startup-file=no`
 end
 
 function get_compat_version(version::VersionNumber, level::String)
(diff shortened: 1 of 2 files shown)


src/PackageCompiler.jl
```diff
@@ -517,19 +516,19 @@ function create_sysimage(packages::Union{Symbol, Vector{Symbol}}=Symbol[];
     # Create the sysimage
     object_file = tempname() * ".o"
     create_sysimg_object_file(object_file, packages;
-                              project=project,
```
Sacha0, line 520: Nice, I forgot about this sugar :).

src/PackageCompiler.jl
```diff
@@ -1064,7 +1061,7 @@ function bundle_artifacts(ctx, dest_dir, library_only; include_lazy_artifacts=tr
                         continue
                     end
                     meta = Pkg.Artifacts.artifact_meta(name, artifacts_toml_path)
-                    meta == nothing && continue
+                    meta === nothing && continue
```
Sacha0, line 1064: Do I remember correctly that `isnothing(meta)` would optimize just as well on 1.7, but not quite yet on 1.6? :)

Sacha0 (approved):
Nice cleanups! :)

I was on the fence about the kwarg sugar when it went in, but seeing it in action is convincing!

src/PackageCompiler.jl
KristofferC: I thin it does, yes.

PR #543, opened 2021-09-17
remove upgrade notes

KristofferC:
I think it is long enough since the remake of PackageCompiler was done that the reference to the old one can be removed.


Sacha0 (approved):
Looks great! :)

PR #546, opened 2021-09-23
Remove support for replacing default sysimage

KristofferC:
This has always been problematic:

- On Windows, it often does not work because it tries to overwrite a file that is already in use.
- On Mac it corrupts the default installation because there is some checksum there.
- It can break tools that expects the default sysimage to behave in a certain way.

The correct way to do it is just to create an alias that starts Julia with your custom sysimage.


davidanthoff:
+1000!!!!

Sacha0 (approved):
Echoing David, 💯!

Sacha0:
Needs a rebase? :)

ViralBShah:
I've definitely been bitten trying to modify the mac installation.

PR #547, opened 2021-09-23
remove warn when detecting dependency on Requires.jl

KristofferC:
Requires has been updated to allow it to work decently with sysimages. Instead of including a file at runtime it now embeds the string in the precompiled file (or sysimage) and evals that string during runtime.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -638,8 +638,6 @@ function restore_default_sysimage()
     return nothing
 end
 
-const REQUIRES = "Requires" => UUID("ae029012-a4dd-5104-9daa-d747884805df")
-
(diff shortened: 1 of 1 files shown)


toollu:
> Requires has been updated to allow it to work decently with sysimages. Instead of including a file at runtime it now embeds the string in the precompiled file (or sysimage) and evals that string during runtime.

On which release of Requires.jl was this fix? The latest release is 1.1.3 but I cannot see the mentioned precompile file embedding in there? Is the change merged yet on Requires.jl?

KristofferC:
https://github.com/JuliaPackaging/Requires.jl/pull/85 is the specific PR.

PR #550, opened 2021-09-23
make `include_lazy_artifacts` default to false

KristofferC:
This really feels like the most sensible default to me. The point of using `lazy` is that you do not want to install the artifacts during package download time and the same is probably true for app build time. Packages can have huge amounts of lazy artifacts that can be unreasonable to bundle.

cc @Sacha0


Sacha0 (approved):
Sounds reasonable to me; thanks Kristoffer! :)

Issue #373, opened 2020-03-23
create_image with packages in Project.toml


bjarthur:
first off, thanks @KristofferC for all your work making PackageCompiler *so* much easier to use!

pardon if i missed this in the docs, but it would be nice if there were a `create_sysimage` method which did not need an explicit list of packages but rather just included all those in the current environment.  as it doesn't seem that Pkg.jl provides a means to get this list (Pkg.status() just prints it out, can't seem to return a list), i resorted to cutting and pasting a manually edited Project.toml onto the command line.  just a thought.  thanks again!

fredrikekre:
See https://github.com/JuliaLang/PackageCompiler.jl/pull/306#issuecomment-596433645

jlperla:
I think the real pain point is manually loading the package file, turning packages into symbols, etc as in https://github.com/julia-vscode/julia-vscode/blob/1da85dc37b54723c1bc4b313d2c11a974c7656b4/scripts/tasks/task_compileenv.jl

So given that, I think maybe a default of all packages is more convenient 

The other thing, missing from that code, is an easy option to use the runtests.jl associated with the project conveniently as a precompile execution file.

KristofferC:
> So given that, I think maybe a default of all packages is more convenient

I agree

adigitoleo:
Thanks for adding this functionality. I have one minor point to raise here.

The other day I was using this thinking that only all _dependencies_ would be added to the sysimage. However, the project itself is also baked in. In other words, in addition to all of the packages listed in the active project's `Project.toml`, the top-level project module itself will be compiled into the sysimage.

I didn't really get that impression from the docs:

> If the packages argument is not passed, all packages **in** the project will be put into the sysimage.

Regardless, it makes this feature awkward to use for my use case: developing a package that has some heavy dependencies that I want to create a sysimage for. Because the project itself is also added in there, when I do something like...

```
julia -JMySysimg.so --project=$PWD -e 'using MyProject; MyProject.main()'
```
...then I am actually running an outdated version of MyProject. So the question is, can the top level project be excluded from the sysimage by default? Or could that be configurable somehow?

On the other hand, I can see an argument here that having a sysimage with frozen dependency packages but not a frozen top-level package could be confusing. But there are practical benefits....

PR #553, opened 2021-09-23
if first argument to `create_sysimage` is left out, include all packages in project

KristofferC:
Fixes https://github.com/JuliaLang/PackageCompiler.jl/issues/373

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -395,10 +395,13 @@ function check_packages_in_project(ctx, packages)
 end
 
 """
-    create_sysimage(packages::Union{String, Vector{String}}; kwargs...)
+    create_sysimage([packages::Union{String, Vector{String}}]; kwargs...)
 
 Create a system image that includes the package(s) in `packages` (given as a
-string or vector).  An attempt to automatically find a compiler will be done but
+string or vector). If the `packages` argument is not passed, all packages in the
+current project will be put into the sysimage.
+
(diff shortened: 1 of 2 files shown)


Sacha0 (approved):
Generally LGTM! Thanks Kristoffer! :)

I do worry a bit that, this being the default, folks might be surprised to find all packages in the project they had live wrapped into the sysimg. But RTM I guess?

KristofferC:
Not sure what you mean. The current functionality when you don't provide the first argument is to give you a sysimage with no new packages in it. I don't think that is really used because its kinda pointless.

Sacha0:
> Not sure what you mean. The current functionality when you don't provide the first argument is to give you a sysimage with no new packages in it. I don't think that is really used because its kinda pointless.

Agreed. In short I just mean it's a change, carry on :).

oscardssmith:
Would it be less surprising if we put this behind a keyword argument (like `full_project=true`) or something?

fredrikekre (approved):

KristofferC:
> Would it be less surprising if we put this behind a keyword argument (like full_project=true) or something?

I don't think so. I think this is the less confusing default and no one is using the current default because it is pointless.

Sacha0:
Looks like this is good to go with a rebase? :)

KristofferC:
Yes, but I want to wait a bit with the breaking PRs until the non-breaking ones are in and out on a release.

PR #557, opened 2021-09-27
fix assumption that Pkg is loadable in a different project

KristofferC:
cc @fredrikekre

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -247,10 +247,6 @@ function run_precompilation_script(project::String, sysimg::String, precompile_f
     return tracefile
 end
 
-function do_ensurecompiled(project, packages, sysimage)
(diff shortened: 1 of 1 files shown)


KristofferC:
Actually, this is not really correct.

KristofferC:
This partially reverts https://github.com/JuliaLang/PackageCompiler.jl/pull/531 so we need to figure out what the issue is there...

KristofferC:
Good, the error is visible in the log:

```
Precompiling project...
  ✓ Zlib_jll
  ✓ MbedTLS_jll
  ✓ LibSSH2_jll
  3 dependencies successfully precompiled in 1 seconds (1 already precompiled)
  4 dependencies errored. To see a full report either run `import Pkg; Pkg.precompile()` or load the packages
```

so now I can investigate.

KristofferC:
I think this should be good now. Errors related to precompilation are gone from the CI log.

PR #559, opened 2021-09-27
prevent accidentally picking up packages outside the project when creating a sysimage

KristofferC:


Sacha0 (approved):
Superficially LGTM; I'm sure I miss any subtleties though :).

PR #563, opened 2021-09-28
share more code between create_app and create_library

KristofferC:
- This now puts the artifacts for `create_app` in the same place as `create_library` (`share/julia/artifacts`).
- `create_app` now also uses the initialization code for the Julia runtime that is provided by `julia_init.c` instead of its own version.


src/embedding_wrapper.c
```diff
@@ -26,35 +21,20 @@ JULIA_DEFINE_FAST_TLS()
...
 
-// main function (windows UTF16 -> UTF8 argument conversion code copied from julia's ui/repl.c)
-int main(int argc, char *argv[])
-{
-    uv_setup_args(argc, argv); // no-op on Windows
-
-    // initialization
-    libsupport_init();
-
-    // Get the current exe path so we can compute a relative depot path
```
KristofferC, line 37: This is actually not needed since the default DEPOT_PATH includes `share/julia`.

KristofferC: Ah, but that only works if one hasn't mucked with the `JULIA_DEPOT_PATH` env var.

src/embedding_wrapper.c
```diff
@@ -26,51 +21,36 @@ JULIA_DEFINE_FAST_TLS()
...
         jl_symbol("PROGRAM_FILE"), (jl_value_t*)jl_cstr_to_string(argv[0]));
 
-    // Set Base.ARGS to `String[ unsafe_string(argv[i]) for i = 1:argc ]`
-    jl_array_t *ARGS = (jl_array_t*)jl_get_global(jl_base_module, jl_symbol("ARGS"));
-    jl_array_grow_end(ARGS, argc - 1);
-    for (int i = 1; i < argc; i++) {
-        jl_value_t *s = (jl_value_t*)jl_cstr_to_string(argv[i]);
-        jl_arrayset(ARGS, s, i - 1);
-    }
+    jl_eval_string("append!(empty!(ARGS), Core.ARGS);");
```
KristofferC, line 42: Julia is easier than C... :man_shrugging:

KristofferC:
I'll split this up into smaller PRs.

PR #570, opened 2021-09-29
Don't warn if modules already loaded during precompilation

IanButterworth:
Given the work is done in a new process, the warnings about packages already being loaded during `Pkg.precompile` are unnecessary.

Fixed in https://github.com/JuliaLang/Pkg.jl/pull/2739 with the introduction of `Pkg.precompile(warn_loaded = false)`, so not in older versions or backported yet.

I used a try catch approach which might be suboptimal.
Is there a better way to check for the presence of a kwarg in a method? Otherwise a julia version check could be done.

For example the error that's thrown in 1.6.3 is this because the kwarg is collected by `kwarg...`

```
julia> Pkg.precompile(warn_loaded = false)
ERROR: type Context has no field warn_loaded
Stacktrace:
 [1] #Context!#40
   @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Pkg/src/Types.jl:427 [inlined]
 [2] precompile(ctx::Pkg.Types.Context; internal_call::Bool, strict::Bool, kwargs::Base.Iterators.Pairs{Symbol, Bool, Tuple{Symbol}, NamedTuple{(:warn_loaded,), Tuple{Bool}}})
   @ Pkg.API /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:921
 [3] #precompile#196
   @ /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:919 [inlined]
 [4] top-level scope
   @ REPL[2]:1
```

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -507,7 +507,15 @@ function create_sysimage(packages::Union{Symbol, Vector{String}, Vector{Symbol}}
     # Need to ensure that Pkg can find the source location of the packages:
     copy!(LOAD_PATH, [project])
     try
-        Pkg.instantiate(ctx, verbose=true, allow_autoprecomp = true)
+        Pkg.instantiate(ctx, verbose=true, allow_autoprecomp = false)
+        try
+            # try to disable warning about precompiling packages that are already loaded given work will be done in new
+            # julia sessions. The `warn_loaded` kwarg was introduced in https://github.com/JuliaLang/Pkg.jl/pull/2739
+            # so isn't present on all versions, so try and do standard precompilation if fails
+            Pkg.precompile(ctx, warn_loaded = false)
+        catch
+            Pkg.precompile(ctx)
+        end
     finally
         copy!(LOAD_PATH, old_load_path)
     end


src/PackageCompiler.jl
```diff
@@ -507,7 +507,14 @@ function create_sysimage(packages::Union{Symbol, Vector{String}, Vector{Symbol}}
     # Need to ensure that Pkg can find the source location of the packages:
     copy!(LOAD_PATH, [project])
     try
-        Pkg.instantiate(ctx, verbose=true, allow_autoprecomp = true)
+        Pkg.instantiate(ctx, verbose=true, allow_autoprecomp = false)
+        if :warn_loaded in Base.kwarg_decl(first(methods(Pkg.precompile, Tuple{Pkg.Types.Context})))
```
KristofferC, line 511: Hm, I also like feature detection but that feels almost feels too much :P. Maybe best to wait for a julia version and just do a regular `VERSION` check?

IanButterworth: the `first(...)` does make this feel a bit too brittle

IanButterworth:
This is the julia commit it landed in https://github.com/JuliaLang/julia/commit/91671dc564d28ea8066c0699a1ed309a06752577

Feel free to take over this PR, otherwise I can get to it next week

PR #598, opened 2021-10-12
Fix the format of Artifacts.toml

zhubonan:
I think that the `i868-w64-mingw32` entry in `Artifacts.toml` should be a table?

Otherwise `Artifacts.artifact_meta` does not filter the platform (something there for backward compatibility?), and this artefact gets downloaded on platforms other than windows, as long as  `include_lazy_artifacts` is `true`. 

https://github.com/JuliaLang/julia/blob/ae8452a9e0b973991c30f27beb2201db1b0ea0d3/stdlib/Artifacts/src/Artifacts.jl#L385-L397

### DIFF
--- a/Artifacts.toml
+++ b/Artifacts.toml
@@ -1,19 +1,19 @@
-[[mingw32]]
+[[mingw]]
 git-tree-sha1 = "fdff308295487f361ef6e8dc2d27f5abe8a6eee9"
 os = "windows"
 arch = "x86_64"
 lazy = true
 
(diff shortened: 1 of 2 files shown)


KristofferC:
Thanks, looks like a typo in https://github.com/JuliaLang/PackageCompiler.jl/commit/522d30cb6b748101b3600689288d4805d5def71d.

zhubonan:
seems to be more complicated, will investigate

PR #631, opened 2021-11-05
Fix code block on apps documentation

guilhermebodin:
The code block was broken on 2.0 docs 

image


KristofferC:
Thanks for catching that!

PR #633, opened 2021-11-11
Fix typo in docs: use correct default cpu target

ikirill:
A very small typo, but it's annoying because it means the user could think they were asked to pass "nativea" instead of "native".


KristofferC:
Heh, thanks.

PR #634, opened 2021-11-11
Update sysimages.md

tetsuji16:
bug fix


KristofferC:
I think the way it is written works fine. Did you test it?

PR #644, opened 2021-12-06
use debug libs when Julia is debug build

jaakkor2:
Julia built with `make debug` does not create `libjulia.so` and `libjulia-internal.so` but `libjulia-debug.so` and `libjulia-internal-debug.so`.


jaakkor2:
x64 on Windows passed tests locally.

PR #649, opened 2021-12-10
add precompile statements for functions called from `main` in apps

JeffBezanson:
The compiler has no other way to know about calls we insert in the `main()` wrapper, so they should have precompile statements. This allows apps to work with the new --strip-ir option.

Remove dependency on at-eval in main --- macros (and eventually maybe the whole front end) might not be available in the system image, so use weaker mechanisms to assign PROGRAM_FILE.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -249,7 +249,8 @@ function create_sysimg_object_file(object_file::String,
                             precompile_statements_file::Vector{String},
                             cpu_target::String,
                             script::Union{Nothing, String},
-                            sysimage_build_args::Cmd)
+                            sysimage_build_args::Cmd,
+                            extra_precompiles::String)
     # Handle precompilation
     precompile_files = String[]
     @debug "running precompilation execution script..."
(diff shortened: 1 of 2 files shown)


KristofferC:
Ah, good point. I used to do this fully with the C API but thought it would be more convenient to just use julia for it. I can go back to the C api if thates better.

JeffBezanson:
Either approach is ok with me.

KristofferC:
If this works well, let's just go with this then.

src/PackageCompiler.jl
```diff
@@ -700,14 +706,24 @@ function create_app(package_dir::String,
     package_name = ctx.env.pkg.name
     project = dirname(ctx.env.project_file)
 
+    # add precompile statements for functions that will be called from the C main() wrapper
+    precompiles = String[]
+    for (_, julia_main) in executables
+        push!(precompiles, "precompile(Tuple{typeof($package_name.$julia_main)})")
```
KristofferC, line 712: CI errors because there is a check to see that a good error message is thrown in case one tries to create an executable from a function that doesn't exist. This could either be guarded with an `isdefined` or throw a good error here in case the function name doesn't exist and fixup the CI check correspondingly.

PR #652, opened 2021-12-14
Allow the user to define arbitrary C variables passed to c_driver_program.

jshap70:
Since PC.jl now exposes `c_driver_program` which allows the users to specify their own C wrapper program, it would be nice to be able to pass in more variables which might be best defined at build time.

For my use case, I would like to bake in a build version at build time. And rather than storing it in julia or needing to compile it into a file, it is far easeier to just have something like:

```C
  #ifdef VERSION
      printf("%s\n", VERSION);
  #else
  ...
```

Where `VERSION` would need to be passed to the compiler like: `-DVERSION="1.3.0"`.

This change will allow for that by passing a `c_defines` dict and letting the user pass in arguments for the compiler to define as a macro.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -672,6 +672,12 @@ compiler (can also include extra arguments to the compiler, like `-g`).
 
 - `sysimage_build_args::Cmd`: A set of command line options that is used in the Julia process building the sysimage,
   for example `-O1 --check-bounds=yes`.
+
+- `c_driver_program`: The C wrapper which serves as an entry point around each target in `executables`. 
+  Defaults to PackageCompiler's embedding_wrapper.c
+
+- `c_defines`: C definitions provided on the CLI for `c_driver_program`. This is only useful for users who are 
+  providing their own `c_driver_program` and wish to provide it definitions at compile time.
 """
 function create_app(package_dir::String,
                     app_dir::String;
@@ -682,6 +688,7 @@ function create_app(package_dir::String,
                     filter_stdlibs::Bool=false,
                     force::Bool=false,
                     c_driver_program::String=String(DEFAULT_EMBEDDING_WRAPPER),
+                    c_defines::Dict{String, String}=Dict{String, String}(),
                     cpu_target::String=default_app_cpu_target(),
                     include_lazy_artifacts::Bool=false,
                     sysimage_build_args::Cmd=``,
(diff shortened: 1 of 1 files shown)


KristofferC:
Can't you already do this by setting an appropriate `JULIA_CC` ENV var?

jshap70:
How would that work? setting it like `ENV["JULIA_CC"] = "/usr/bin/gcc -DVERSION=1.0"`? 

it does seem like the way the variable is handled in `run_compiler` would allow for that, however at least to me that feels very unintuitive. I not only would not have known to try it, but also to me that feels more like an option that explicitly is "the location of the compiler binary" and not "the beginning of the compile command". 

Further, if I'm on windows I really want to leave the logic alone for using gcc from the local mingw. Otherwise I'm hard-coding a magic directory string for where it will be or adding the dependency to my own project.

KristofferC:
Okay, what I don't want is to add a zillion of keyword arguments for all kind of compiler features. There should in my opinion be one at most one argument for advanced users where they can customize the compiler options.

Also, the docs do say:

> An attempt to automatically find a compiler will be done but can also be given
explicitly by setting the environment variable `JULIA_CC` to a path to a
compiler (can also include extra arguments to the compiler, like `-g`).

jshap70:
very fair. and I must have just missed that in the docs, however I still feel like that makes those options inaccessible on windows. 

As a long term solution, why not organize the compiler flags and such into some sort of organized object/struct, where all options like binary location, environment, flags, cpu target, etc are stored. The current functions as is could just keep complexity away from the user and simply construct it in place and then delegate off to a version which accepts the struct, but an advanced user could manage their own versions of it if they wanted and then call directly into the versions which use the struct?

jshap70:
how would you feel if instead this looked something like: 

```diff
diff --git a/src/PackageCompiler.jl b/src/PackageCompiler.jl
index 4068520..4d0dbb1 100644
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -118,6 +118,7 @@ const WARNED_CPP_COMPILER = Ref{Bool}(false)

 function run_compiler(cmd::Cmd; cplusplus::Bool=false)
     cc = get(ENV, "JULIA_CC", nothing)
+    cc_flags = get(ENV, "JULIA_CC_FLAGS", "")
     path = nothing
     @static if Sys.iswindows()
         path = joinpath(LazyArtifacts.artifact"mingw-w64", (Int==Int64 ? "mingw64" : "mingw32"), "bin", cplusplus ? "g++.exe" : "gcc.exe")
@@ -158,7 +159,7 @@ function run_compiler(cmd::Cmd; cplusplus::Bool=false)
     if path !== nothing
         compiler_cmd = addenv(compiler_cmd, "PATH" => string(ENV["PATH"], ";", dirname(path)))
     end
-    full_cmd = `$compiler_cmd $cmd`
+    full_cmd = `$compiler_cmd $cc_flags $cmd`
     @debug "running $full_cmd"
     run(full_cmd)
 end
```

jshap70:
Closed in favor of #665

PR #659, opened 2022-01-09
(fix) ERROR: expected a Cint return value from fun

mroavi:


KristofferC:
I would have thought the return type annotation on the function would have handled this.

mroavi:
Hmm. I just got this error while testing this example. 

```julia
Example.domath(5) = 10
sin(0.0) = 0.0
outputo: ok
myrand: ok
n = 20000000
8
      From worker 5:    8
      From worker 2:    8
      From worker 3:    8
      From worker 4:    8
ERROR: TypeError: in typeassert, expected Int32, got a value of type Int64
Stacktrace:
 [1] real_main()
   @ MyApp ~/Dropbox/TUe/PhD/software/julia/PackageCompiler/creating-an-app/MyApp/src/MyApp.jl:81
 [2] julia_main()
   @ MyApp ~/Dropbox/TUe/PhD/software/julia/PackageCompiler/creating-an-app/MyApp/src/MyApp.jl:20
 [3] top-level scope
   @ none:1
ERROR: expected a Cint return value from function MyApp.julia_main
```

Maybe you could run it in your machine to see if you can reproduce it.

PR #665, opened 2022-02-12
Allow the user to pass arbitrary C flags (on all platforms)

julie-is-late:
Since PC.jl now exposes `c_driver_program` which allows the users to
specify their own C wrapper program, it would be nice to be able to pass
in more variables which might be best defined at build time.

Currently this functionality is effectively limited to Linux as
specifying $JULIA_CC will prevent PC.jl from inferring gcc.exe from
the relative mingw installation.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -118,6 +118,7 @@ const WARNED_CPP_COMPILER = Ref{Bool}(false)
 
 function run_compiler(cmd::Cmd; cplusplus::Bool=false)
     cc = get(ENV, "JULIA_CC", nothing)
+    cc_flags = get(ENV, "JULIA_CC_FLAGS", nothing)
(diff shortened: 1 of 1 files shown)


jshap70:
Replaces #652 to no longer require a keyword argument.

jshap70:
Apologies for the commit spam. Sleepy brain forgot to activate the project before running the tests and was very confused about why it was failing remotely.

sjkelly:
Also see:
https://github.com/JuliaLang/PackageCompiler.jl/pull/920

In https://github.com/JuliaLang/PackageCompiler.jl/pull/929 I am attempting to make a config struct, which seems like a more composable and streamlined way to to this.

julie-is-late:
That’s a much better way to organize it in general, so having this covered in that would be fantastic.

PR #666, opened 2022-02-14
Update minimum tested Julia version to ~1.6.0

kmsquire:
* Previously, this was set to '1', which (presumably) tested the
  latest stable release.
* We currently promise to support 1.6, so we should always test this,
  even if the latest stable release is bigger


KristofferC:
This is fixed

PR #669, opened 2022-02-28
Update README.md

nlw0:
Couple of typos. Not sure about the second one, the previous sentence was "from of Julia Code", and it was changed to "form of Julia Code", but I have a feeling the correct should be "from Julia Code", does that make sense?


README.md
```diff
@@ -10,7 +10,7 @@ PackageCompiler is a Julia package with three main purposes:
 
   2. Creating "apps" which are a bundle of files including an executable that can be sent and run on other machines without Julia being installed on that machine.
 
-  3. Creating a relocatable C library bundle form of Julia code.
```
sjkelly, line 13: Actually this seems correct.

nlw0: I still don't understand what it means but it must be me then :+1:

sjkelly: At first I thought it was not clear either, but a second read it made sense. Maybe inversion helps clarity:

"Creating a relocatable bundle of Julia code in the form of a C library."

PR #672, opened 2022-03-14
`build` and then `precompile` in `ensurecompiled`

ven-k:
- In certain cases like when `Conda` is invloved, the `ensurecompiled` fails if Conda isn't built properly.


KristofferC:
I would have thought https://github.com/JuliaLang/PackageCompiler.jl/blob/6a3bdcc13743ba5e94c814162a0b8d30a720eacb/src/PackageCompiler.jl#L450 would have built packages if they got downloaded.

ven-k:
It's been a hit and miss. Most of the time it isn't an issue. But couple of times Conda wasn't built properly and resulted in failed `create_sysimage`

PR #673, opened 2022-03-15
Allow certain dependencies to be excluded for sysimg

### LINKED ISSUE #639: Lazy downloading artifacts with custom sysimage
Some images created by `create_sysimage()` tries downloading artifacts after it's loaded first time.

```
$ julia
julia> using PackageCompiler

julia> create_sysimage([:Gadfly], sysimage_path="gadfly.dylib")
✔ [01m:48s] PackageCompiler: compiling incremental system image

$ julia -Jgadfly.dylib
 Downloading artifact: MKL
    Downloading [================>                        ]  39.2 %
```

I believe this issue started happening since recent versions of PackageCompiler, presumably from 1.6 or 1.7. I'm not entirely sure if there was any relevant change in the said package, i.e. MKL.jl, triggering this issue.

I saw `include_lazy_artifacts` option was recently added to `create_app()` and `create_library()`, but there seems to be no equivalent option for `create_sysimage()`.

```
julia> versioninfo()
Julia Version 1.6.4
Commit 35f0c911f4 (2021-11-19 03:54 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.5.0)
  CPU: Intel(R) Core(TM) i7-10700K CPU @ 3.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
```

PackageCompiler.jl used in the example above was v2.0.2.

zhubonan:
It is possible for a package to depend on more than it needs, for eaxmple, to allow switching backends at run time. One example is the [FFTW.jl](https://github.com/JuliaMath/FFTW.jl). 

In some cases, it is desirable to exclude certain dependencies to avoid downloading or shipping with shipping artifacts that are not needed for an app (e.g. the [MKL_jll.jl](https://github.com/JuliaBinaryWrappers/MKL_jll.jl/blob/main/Artifacts.toml) which FFTW.jl depends on, but not used by default).

Potentially fix #639, where the unused package with lazy artifact can be explicitly excluded for building system image/app.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -391,6 +391,9 @@ compiler (can also include extra arguments to the compiler, like `-g`).
    transitive dependencies into the sysimage. This only makes a difference if some
    packages do not load all their dependencies when themselves are loaded. Defaults to `true`.
 
+- `excluded_dependencies`: A Vector of names of the (tansitive) dependencies to be excluded
+  from the system image.
+
 ### Advanced keyword arguments
 
 - `base_sysimage::Union{Nothing, String}`: If a `String`, names an existing sysimage upon which to build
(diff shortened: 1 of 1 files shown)


groverburger:
This feature looks great! I was just thinking of adding something like this myself. I would love to see this merged in soon specifically for excluding Revise.jl during package compilation. Any news on that front from the maintainers?

One nit: the docstring refers to the kwarg as `excluded_dependencies` when it is called `excluded_transitive_dependencies`.

PR #693, opened 2022-06-02
In the `create_sysimage` function, allow the `project` kwarg to take either the directory of the project or the filename of the `Project.toml` file

DilumAluthge:


DilumAluthge:
I don't like this approach. It's cleaner to require the user to provide the directory. I'll just update the docstring to clarify that `project` should be a directory.

Issue #691, opened 2022-05-26
PackageCompiler does not download artifacts for MKL now


jsjie:
after [this commt](https://github.com/JuliaLang/PackageCompiler.jl/commit/1cc1cdf7063b73718013a3f613f29f98045c6f6e)
in `bundle_artifacts`, newly added if clause behave slightly differently, and cannot find certain artifacts, in my case, MKL and OpenMPI.
test in:
```
julia> platform = Base.BinaryPlatforms.HostPlatform()
Linux x86_64 {cxxstring_abi=cxx11, julia_version=1.7.1, libc=glibc, libgfortran_version=5.0.0, libstdcxx_version=3.4.29}
```

lucasmsoares96:
I'm having this same problem.

```
julia> Base.BinaryPlatforms.HostPlatform()
Linux x86_64 {cxxstring_abi=cxx11, julia_version=1.9.0, libc=glibc, libgfortran_version=5.0.0, libstdcxx_version=3.4.30}
```

```
julia> versioninfo()
Julia Version 1.9.0-rc2
Commit 72aec423c2a (2023-04-01 10:41 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, haswell)
  Threads: 4 on 8 virtual cores
Environment:
  JULIA_NUM_THREADS = 4
```

PR #696, opened 2022-06-13
fix lazy artifacts

KristofferC:
Should fix https://github.com/JuliaLang/PackageCompiler.jl/issues/691


KwatMDPhD:
I hope this pull request merges. I'm getting the same MKL error. Thanks @KristofferC

PR #697, opened 2022-06-14
Add a check for a package's __init__ being called appropriately

IanButterworth:
We're seeing odd behavior with `__init__()` in a dep being called multiple times during sysimage precompilation.

Just adding checks to see if it's reproducible here.

cc. @octogonapus


KristofferC:
(1.6 is known to fail)

IanButterworth:
The nightly failures are on master too

IanButterworth:
Ok, our issue isn't replicated here, but perhaps these tests are worth adding?

KristofferC:
It just seem so arbitrary. Until there is something showing that there is something going on with the init functions and that this has something to do with PackageCompiler, I'm a bit hesitant in adding this.

IanButterworth:
ok! We're still trying to figure out what's happening.

I thought PackageCompiler took over when `__init__()`s were run, so thought guaranteeing that they're run exactly once (which the julia docs guarantee) was worth testing. Or perhaps the historical issues I'm vaguely thinking of were fixed in base.

Octogonapus:
@IanButterworth This turned out to not be an issue, I think we should close this PR

PR #705, opened 2022-07-31
check if manifest file exists instead of if manifest empty

simonbyrne:
Theoretically you could compile a package with an empty manifest


simonbyrne:
Note that 1.6 tests are already failing (#706)

src/PackageCompiler.jl
```diff
@@ -63,7 +63,7 @@ function create_pkg_context(project)
         error("could not find project at $(repr(project))")
     end
     ctx = Pkg.Types.Context(env=Pkg.Types.EnvCache(project_toml_path))
-    if isempty(ctx.env.manifest)
+    if isfile(ctx.env.manifest_file)
```
KristofferC, line 66: `!`.. ?

PR #711, opened 2022-08-08
also bundle certificate

KristofferC:
should fix https://github.com/JuliaLang/MbedTLS.jl/pull/219#issuecomment-1206080455


KwatMDPhD:
Works.

KwatMDPhD:
🙌

PR #712, opened 2022-08-08
Add concurrency control to GitHub workflow runs

davidanthoff:
If this works the way I think it works (and I've tested it a bit on another repo) then this would mean that if one pushes a new commit to a PR or branch that triggers a workflow run, any previous still running instance of the same workflow for that same PR or branch would be cancelled.

Should overall speed things up because one should run less often into resource constraints.


KristofferC:
Sounds good.

Issue #706, opened 2022-08-01
Tests are failing on Julia 1.6


simonbyrne:
https://github.com/JuliaLang/PackageCompiler.jl/runs/7175755312?check_suite_focus=true#step:4:340
```
InitError(mod=:LLVMExtra_jll, error=ErrorException("Cannot locate artifact 'LLVMExtra' in '/tmp/jl_s8rPQE/packages/LLVMExtra_jll/RuRRc/Artifacts.toml'"))
```

KristofferC:
Yeah, I merged a test that failed on 1.6 at a time when we didn't have 1.6 CI. It is a bit tricky to make this conditional on the julia version though.

DilumAluthge:
Should be fixed by #771

PR #731, opened 2022-10-14
Factor out the "find compiler" logic into a separate `get_compiler_cmd` function

DilumAluthge:


DilumAluthge:
The CI failures on Julia 1.6 are #706

Issue #739, opened 2022-11-16
Provide more useful info when sysimage generation fails


IanButterworth:
There's very little to go off here (which is frustrating after a long build process)
```
✖ [13m:52s] PackageCompiler: compiling incremental system image
ERROR: failed process: Process(`/home/ian/.julia/juliaup/julia-1.8.2+0.x64/bin/julia --color=yes --startup-file=no --cpu-target=native -O3 --sysimage=/home/ian/.julia/juliaup/julia-1.8.2+0.x64/lib/julia/sys.so --project=/home/ian/.julia/packages/Foo/A61yO --output-o=/tmp/jl_j0by5sNSt8.o /tmp/jl_ZurqZanGvy`, ProcessSignaled(9)) [0]

Stacktrace:
  [1] pipeline_error
    @ ./process.jl:565 [inlined]
  [2] run(::Cmd; wait::Bool)
    @ Base ./process.jl:480
  [3] run
    @ ./process.jl:477 [inlined]
  [4] #15
    @ ~/.julia/packages/PackageCompiler/DKijm/ext/TerminalSpinners.jl:157 [inlined]
  [5] spin(f::PackageCompiler.var"#15#16"{Cmd}, s::PackageCompiler.TerminalSpinners.Spinner{Base.TTY})
    @ PackageCompiler.TerminalSpinners ~/.julia/packages/PackageCompiler/DKijm/ext/TerminalSpinners.jl:164
  [6] macro expansion
    @ ~/.julia/packages/PackageCompiler/DKijm/ext/TerminalSpinners.jl:157 [inlined]
  [7] create_sysimg_object_file(object_file::String, packages::Vector{String}, packages_sysimg::Set{Base.PkgId}; project::String, base_sysimage::String, precompile_execution_file::Vector{String}, precompile_statements_file::Vector{String}, cpu_target::String, script::Nothing, sysimage_build_args::Cmd, extra_precompiles::String, incremental::Bool)
    @ PackageCompiler ~/.julia/packages/PackageCompiler/DKijm/src/PackageCompiler.jl:364
[...]

PR #740, opened 2022-11-17
Monitor for low free memory during sysimage generation

IanButterworth:
Helping with #739 

In lieu of any other way to gracefully show that the subprocess has been OOM killed, this warns once if it's likely


IanButterworth:
This seems like a conservative change only in the error path, and has proven helpful to our real-world usage, so I will merge later if I don't hear otherwise @KristofferC

PR #750, opened 2022-12-28
suggest using Plots.__init__()

rdiaz02:
This solves issue in https://discourse.julialang.org/t/plots-do-not-display-unless-i-call-gui-when-using-custom-sysimage/92207


KristofferC:
But `Plots.__init__` should run when the sysimage is loaded. Does something else overwrite what it does?

rdiaz02:
I don't know.  But following the example in the documentation (https://julialang.github.io/PackageCompiler.jl/dev/examples/plots.html) definitely does not work for me now, as reported in https://discourse.julialang.org/t/plots-do-not-display-unless-i-call-gui-when-using-custom-sysimage/92207 . I've checked this in two different machines (both Linux, Debian).



These are the steps (the same as in  https://discourse.julialang.org/t/plots-do-not-display-unless-i-call-gui-when-using-custom-sysimage/92207):

1. Create file ` precompile_plots.jl` that contains

```julia-repl
using Plots
p = plot(rand(2,2))
display(p)
```

2.  Start Julia: `julia --startup-file=no`

3.  Type 

```julia-repl
using PackageCompiler
create_sysimage(["Plots"], sysimage_path="sys_plots.so", precompile_execution_file="precompile_plots.jl")
```

4. Start Julia as

`julia --startup-file=no --sysimage sys_plots.so`


5. In the REPL type

```
using Plots
plot(rand(1))
```

The only thing shown is

```
Plot{Plots.GRBackend() n=1}
```
I have no idea what could be overwriting what `Plots.__init__` does. But unless I do `Plots.__init__()`  , I need to do `gui()` (or `gui(plot(rand(1)))` or `plot(rand(1), show = true)`) for the plot to be displayed.

Note also that a similar thing seems to happen with Makie, as reported in https://discourse.julialang.org/t/makie-doesnt-display-plot-when-using-a-custom-julia-sysimage/38515/4 .

ufechner7:
Why are the tests failing? Can this be merged?

Issue #755, opened 2023-01-06
RNG not being seeded during precompilation


IanButterworth:
Maybe this needs to be moved to julia issues.

On nightly the following is failing
https://github.com/JuliaLang/PackageCompiler.jl/blob/8d71b02705a4ceae8ca1481d09107938f68b0d57/examples/MyApp/src/MyApp.jl#L73-L74

where myrand is a const in the module
https://github.com/JuliaLang/PackageCompiler.jl/blob/8d71b02705a4ceae8ca1481d09107938f68b0d57/examples/MyApp/src/MyApp.jl#L14

See https://github.com/JuliaLang/PackageCompiler.jl/actions/runs/3857713317/jobs/6575476152

@vchuravy is this known?

IanButterworth:
I guess https://github.com/JuliaLang/PackageCompiler.jl/pull/756 would fix this in PackageCompiler, but it might be an upstream bug?

IanButterworth:
I can't reproduce this in basic julia master without a custom sysimage of the app
```
julia> using MyApp
[ Info: Precompiling MyApp [f943f3d7-887a-4ed5-b0c0-a1d6899aa8f5]
  Downloaded artifact: MKL

julia> MyApp.myrand
0.6190234895057138
```
```
julia> using MyApp

julia> MyApp.myrand
0.6190234895057138
```

KristofferC:
This indicates to me that the precompile file of MyApp is not used but instead that the `using MyApp` in the process that creates the sysimage is getting compiled there. So we could maybe add `ENV["JULIA_DEBUG"]` to the process creating the sysimage and see if the ji file indeed gets loaded or if it gets rejected for some reason.

IanButterworth:
Ok, pushed to https://github.com/JuliaLang/PackageCompiler.jl/pull/757 Feel free to push etc.

IanButterworth:
Setting `ENV["JULIA_DEBUG"]="loading"` in #757 shows the cache files are being rejected, for instance
```
┌ Debug: Rejecting cache file /tmp/jl_aCrjdm/compiled/v1.9/LibSSH2_jll/K6mup_TmZuG.ji for LibSSH2_jll [29816b5a-b9ab-546f-933c-edad1886dfa8] since the flags are mismatched
│   cachefile_flags = 35
│   current_flags = 51
└ @ Base loading.jl:2526
```

It'd be good for that log to be more informative

PR #756, opened 2023-01-06
Disable pkgimages

IanButterworth:
I don't believe PackageCompiler wants pkgimages at all in the subprocess work, so disable them

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,12 +35,13 @@ jobs:
           - os: macOS-latest
             julia-arch: x86
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


KristofferC:
This isn't 100% clear to me. First, we run a `Pkg.precompile` before doing the sysimage stuff but that will be worthless if we don't use the results of that.

And most people will probably have their environment already precompiled with pkg images. If this turns it off, you have to precompile everything from scratch.

IanButterworth:
I was just thinking of it as taking more time given the pkgimages image generation step, but I guess you're right.

There's two test failures to tackle then. One is https://github.com/JuliaLang/PackageCompiler.jl/issues/755 the other is that outputo returns a file when pkgimages is enabled, which tests don't expect

Issue #677, opened 2022-04-06
Fail early if no compiler can be found?


baggepinnen:
If possible, it would perhaps be nice to detect the presence of a compiler before spending the time to build the system image?
```
✔ [07m:05s] PackageCompiler: compiling incremental system image
ERROR: LoadError: could not find a compiler, looked for g++, clang++, gcc and clang
```

Issue #727, opened 2022-10-03
Check c++ compiler earlier


KronosTheLate:
I just generated a sysimage, taking a lot of time, only to fail in the end because I was missing a c++ compiler.
Could a simple check be made in the start to avoid wasting time for users without a c++ compiler installed?

Issue #716, opened 2022-08-21
Split `run_compiler` so users can get the location of the compiler?


jlapeyre:
I want to use `PackageCompiler` for a completely different project that involves compiling C code. It would be useful if I could just find the compiler. You could rewrite `run_compiler` like this:
```julia
function run_compiler(cmd::Cmd; cplusplus::Bool=false)
    compiler_cmd = find_compiler_cmd(cplusplus)
    full_cmd = `$compiler_cmd $cmd`
    @debug "running $full_cmd"
    run(full_cmd) 
end
```
And put the rest of the existing function body in `find_compiler_cmd` instead. Then we could call `find_compiler_cmd` for other projects.

As it is, we can still just call `run_compiler`. But, if the function were split as suggested here, we'd have more flexibility, and options for debugging and handling exceptional situations.

This might help resolving #714 . For example, you could call `find_compiler_cmd` before trying to build your project.

DilumAluthge:
Closed by #731

DilumAluthge:
#731 introduced `get_compiler_cmd`, which is equivalent to the hypothetical `find_compiler_cmd` function in the OP.

Issue #714, opened 2022-08-10
Request: Warning for lack of appropriate compiler 


bclyons12:
I recently tried to compile an app using 
```
PackageCompiler.create_app(package_dir, target_dir, force = true, precompile_execution_file = build_dir * "/generate_precompile.jl")
```
but without loading an appropriate version of gcc. After creating the system image, I got several hundred lines of errors. I can't include it all (too long for a Issue boy apparently), but here's a sampling:

<details><summary>Expand for output</summary><p>

```
[lyonsbc@irisa]% julia build/build.jl /fusion/projects/codes/qed/v1.2.0
  Activating project at `~/.julia/dev/QED`
┌ Warning: it is not recommended to create an app/library without a preexisting manifest
└ @ PackageCompiler ~/.julia/packages/PackageCompiler/KhQCf/src/PackageCompiler.jl:67
PackageCompiler: bundled artifacts:
  └── OpenSpecFun_jll - 656.279 KiB
  Total artifact file size: 656.279 KiB
┌ Warning: it is not recommended to create an app/library without a preexisting manifest
└ @ PackageCompiler ~/.julia/packages/PackageCompiler/KhQCf/src/PackageCompiler.jl:67
✔ [04m:02s] PackageCompiler: compiling base system image (incremental=false)
Precompiling project...
  42 dependencies successfully precompiled in 233 seconds
[ Info: PackageCompiler: Executing /home/lyonsbc/.julia/dev/QED/build/generate_precompile.jl => /tmp/jl_packagecompiler_suPJ9Y/jl_sdofMu
[ Info: PackageCompiler: Done
✔ [07m:48s] PackageCompiler: compiling nonincremental system image
In file included from /fusion/projects/omfit-results/lyonsbc/julia-1.7.3/include/julia/libsupport.h:10,
                 from /fusion/projects/omfit-results/lyonsbc/julia-1.7.3/include/julia/julia.h:13,
                 from /home/lyonsbc/.julia/packages/PackageCompiler/KhQCf/src/embedding_wrapper.c:7:
/fusion/projects/omfit-results/lyonsbc/julia-1.7.3/include/julia/dtypes.h: In function ‘jl_load_ptraligned_f64’:
/fusion/projects/omfit-results/lyonsbc/julia-1.7.3/include/julia/dtypes.h:335: warning: implicit declaration of function ‘__builtin_assume_aligned’
[...]

PR #769, opened 2023-01-26
run `get_compiler_cmd` earlier, to error if not found

sjkelly:
Closes: #677, #714, #716, #727

In many scenarios a default user shell may not have access to a compiler. The current behavior runs the checks after compiling the sysimg, which may take many minutes. This moves the check earlier so that this error happens earlier. It also splits the `cmd` and `args` such that it may be re-used or potentially altered by a user.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -184,9 +184,8 @@ function get_compiler_cmd(; cplusplus::Bool=false)
     return compiler_cmd
 end
 
-function run_compiler(cmd::Cmd; cplusplus::Bool=false)
-    compiler_cmd = get_compiler_cmd(; cplusplus)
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Why is the diff so big here? I think this should be a very small PR, right? Just call `get_compiler_cmd()` at the beginning of `create_sysimage` and `create_app`.

sjkelly:
Yeah, I suppose that way is simpler... You are doing the work twice, but it is simple work.

DilumAluthge:
Yeah, it's relatively simple work, and it doesn't take very long to do (especially relative to how long building the sysimage will take), so I think it's okay to do that work twice.

DilumAluthge:
Closing in favor of #770

PR #770, opened 2023-01-26
Call `get_compiler_cmd()` at the beginning of `create_sysimage` and `create_app`

### LINKED ISSUE #677: Fail early if no compiler can be found?

### LINKED ISSUE #727: Check c++ compiler earlier

DilumAluthge:
That way, users will immediately get an error if a compiler cannot be found, instead of having to wait a long time before the error is thrown.

Replaces #769
Closes #769 
Closes #677 
Closes #727

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -483,6 +483,7 @@ function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector
                          compat_level::String="major",
                          extra_precompiles::String = "",
                          )
+    get_compiler_cmd()
 
     if filter_stdlibs && incremental
         error("must use `incremental=false` to use `filter_stdlibs=true`")
@@ -748,6 +749,7 @@ function create_app(package_dir::String,
                     sysimage_build_args::Cmd=``,
                     include_transitive_dependencies::Bool=true)
     warn_official()
+    get_compiler_cmd()
 
     ctx = create_pkg_context(package_dir)
     ctx.env.pkg === nothing && error("expected package to have a `name`-entry")
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -483,7 +483,8 @@ function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector
                          compat_level::String="major",
                          extra_precompiles::String = "",
                          )
-    # We call this at the very beginning to make sure that the user has a compiler available. Therefore, if no compiler is fiund, we throw an error immediately, instead of making the user wait a while before the error is thrown.
+    # We call this at the very beginning to make sure that the user has a compiler available. Therefore, if no compiler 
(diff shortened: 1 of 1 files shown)


DilumAluthge:
I'm not sure what the current state of CI is.

@IanButterworth Are these CI failures expected?

src/PackageCompiler.jl
```diff
@@ -483,6 +483,7 @@ function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector
                          compat_level::String="major",
                          extra_precompiles::String = "",
                          )
+    get_compiler_cmd()
```
IanButterworth, line 486: Perhaps add a comment to explain why this is called in both locations but not used?

IanButterworth:
1.6 won't pass. The rest should but there are gremlins

src/PackageCompiler.jl
DilumAluthge: Done!

src/PackageCompiler.jl
```diff
@@ -483,6 +483,8 @@ function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector
                          compat_level::String="major",
                          extra_precompiles::String = "",
                          )
+    # We call this at the very beginning to make sure that the user has a compiler available. Therefore, if no compiler is fiund, we throw an error immediately, instead of making the user wait a while before the error is thrown.
```
IanButterworth, line 486: typo on found and linewidth
```suggestion
    # We call this at the very beginning to make sure that the user has a compiler available.  Therefore, if no compiler 
    # is found, we throw an error immediately, instead of making the user wait a while before the error is thrown.
```

src/PackageCompiler.jl
```diff
@@ -748,6 +750,8 @@ function create_app(package_dir::String,
                     sysimage_build_args::Cmd=``,
                     include_transitive_dependencies::Bool=true)
     warn_official()
+    # We call this at the very beginning to make sure that the user has a compiler available. Therefore, if no compiler is fiund, we throw an error immediately, instead of making the user wait a while before the error is thrown.
```
IanButterworth, line 753:
```suggestion
    # We call this at the very beginning to make sure that the user has a compiler available.  Therefore, if no compiler 
    # is found, we throw an error immediately, instead of making the user wait a while before the error is thrown.
```

IanButterworth (approved):

PR #772, opened 2023-01-27
use smaller micromamba_jll rather than huge MKL_jll to test lazy artifacts

### LINKED ISSUE #766: Avoid redownloading MKL_jll for each testset
Because each test starts with a clean depot state, as I think it should, they each download MKL_jll, which can make tests a lot slower (especially on my wifi which is relatively mediocre).

It'd be nice to figure out a way to ensure that MKL_jll is downloaded appropriately for apps, but avoid downloading it each time.

IanButterworth:
Fixes #766 

MKL_jll was used to test lazy artifact downloading. 

micromamba_jll artifacts are also lazy but a lot smaller (~5-7 MiB) than MKL_jll which is nearly 1GB

This switches over. 

https://github.com/JuliaBinaryWrappers/micromamba_jll.jl/blob/main/Artifacts.toml


DilumAluthge:
> micromamba_jll artifacts are lazy but a lot smaller (~5-7 MiB) than MKL_jll which is nearly 1GB

To provide context for others (like me) that might be confused by this sentence, the MKL_jll artifacts are lazy (https://github.com/JuliaBinaryWrappers/MKL_jll.jl/blob/main/Artifacts.toml).

PR #774, opened 2023-01-27
RFC: propagate build flags into sysimg

sjkelly:
This removes the enforced `-O` and `g` levels,
and propagates the passed arguments to the entire build. We filter a few arguments from the base sysimg to prevent some failures and avoid known bugs.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -211,7 +211,7 @@ function rewrite_sysimg_jl_only_needed_stdlibs(stdlibs::Vector{String})
         r"stdlibs = \[(.*?)\]"s => string("stdlibs = [", join(":" .* stdlibs, ",\n"), "]"))
 end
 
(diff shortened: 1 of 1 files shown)


src/PackageCompiler.jl
```diff
@@ -211,25 +211,35 @@ function rewrite_sysimg_jl_only_needed_stdlibs(stdlibs::Vector{String})
         r"stdlibs = \[(.*?)\]"s => string("stdlibs = [", join(":" .* stdlibs, ",\n"), "]"))
 end
 
-function create_fresh_base_sysimage(stdlibs::Vector{String}; cpu_target::String)
+function create_fresh_base_sysimage(stdlibs::Vector{String}; cpu_target::String, sysimage_build_args)
```
KristofferC, line 214: Please add a type for this argument.

PR #780, opened 2023-02-02
ensure threading is disabled where not allowed

IanButterworth:
It seems like threading might not be allowed/expected on 1.9+ in some of these stages, which was causing #778. 
@vtjnash is that right?

This fixes/patches https://github.com/JuliaLang/PackageCompiler.jl/issues/778 in cases where the user has `JULIA_NUM_THREADS` set to > 1

Testing locally, this fixes it in the sense that CPU load goes down to 100% per precompile task.

However, precompilation still takes a lot longer than it should. It seems like the 1 active thread is spinning or something


IanButterworth:
This is the wrong approach because threads should always respect what the user requested

JeffBezanson:
I don't think this is necessarily wrong --- we can certainly choose how many threads to use in a process that we spawn for the purpose of building a system image.

IanButterworth:
As discussed, a proposal to fix this upstream https://github.com/JuliaLang/julia/pull/48592

PR #791, opened 2023-02-18
Add excluded_packages keyword argument for create_sysimage

zhubonan:
Added a `excluded_packages` keyword to `create_sysimage` which allows certain packages to be unconditionally excluded in the sysimage. 

Such option can be useful when one want to compile a sysimage for package development where the dependences contain heavy packages with long load times (e.g. found using `@time_imports using MyDevPackage` in Julia >= 1.8). 

In this case, building a system image including the dependencies can significantly reduce the load times of the package in development. 

For example, one may want to create a system image with `MyDevPackage` included in the `packages` keyword argument with `include_transitive_dependencies=true` to ensure its dependencies are included. Then put `MyDevPackage` inside `excluded_packages` so `MyDevPackage` itself is not included in the sysimage (otherwise one cannot develop `MyDevPackage`.


sloede:
Are there any updates for this PR? To have an easy way for selecting all dependencies of a package without the package itself seems like a nice improvement for usability

zhubonan:
> Are there any updates for this PR? To have an easy way for selecting all dependencies of a package without the package itself seems like a nice improvement for usability

No update from my side - waiting for inputs from the maintainer I guess.

Issue #768, opened 2023-01-25
Incremental sysimage build doesn't support packages with same name but different uuid


krynju:
This one is similiar to #767, but setup and place where it errors are different.

## Setup

Simple two projects that have a top level dependency PackageA, but with different uuids

```julia
using PackageCompiler

using Pkg

Pkg.activate("project1") # has Package1 with uuid=1234
create_sysimage(["Package1"]; sysimage_path="sysimage5.so")

Pkg.activate("project2") # has Package1 with uuid=5678
create_sysimage(["Package1"]; sysimage_path="sysimage6.so", base_sysimage="sysimage5.so")
```

## Error

```julia
julia> create_sysimage(["Package1"]; sysimage_path="sysimage6.so", base_sysimage="sysimage5.so")
⠙ [00m:07s] PackageCompiler: compiling incremental system imageLoadError
⡆ [00m:08s] PackageCompiler: compiling incremental system imageimporting Package1 into Main conflicts with an existing identifier
in expression starting at /tmp/jl_bOn2oN:7
✖ [00m:08s] PackageCompiler: compiling incremental system image
ERROR: failed process: Process(`/home/krynju/julia-1.7.3/bin/julia --color=yes --startup-file=no --cpu-target=native -O3 --sysimage=sysimage5.so --project=/home/krynju/sameuuidtest/project2 --output-o=/tmp/jl_sKRDQC.o /tmp/jl_bOn2oN`, ProcessExited(1)) [1]
```

Error appears due to this code: https://github.com/JuliaLang/PackageCompiler.jl/blob/05f6854fb9147bd0faae7b442555cddc69457f64/src/PackageCompiler.jl#L377-L382

## Solution

I worked around this by commenting out this part of code and everything looks to be functioning just right. 

The error didn't appear if the package was the same uuid, but a different version, which is probably something weird to handle as well

A build flag to turn off this import behavior would probably be good enough as I'm not sure what the import to Main is dedicated for

KristofferC:
> A build flag to turn off this import behavior would probably be good enough as I'm not sure what the import to Main is dedicated for

It is there to be able to get a reference to the module (e.g. `using .Plots`) without having to have a Project file available to lookup the UUID when doing `using Plots`.

staticfloat:
Here's a simple way to reproduce this:

```
julia> versioninfo()
Julia Version 1.9.0-rc1
[...]

Issue #767, opened 2023-01-24
Precompile stage doesn't support packages with the same names and different uuids


krynju:
## Setup
In a setup like this:

1. PackageA (depends on PackageC with uuid 1234)
2. PackageB (depends on PackageC with uuid 5678)

The `loaded_modules` will look something like this:
```
julia> Base.loaded_modules
Dict{Base.PkgId, Module} with 48 entries:
  PackageC [1234]           => Package1
  PackageC [5678]           => Package1
```

## Error
This setup fails at this line, because it hits a constant redefinition error 
https://github.com/JuliaLang/PackageCompiler.jl/blob/05f6854fb9147bd0faae7b442555cddc69457f64/src/PackageCompiler.jl#L320

```
⣠ [00m:08s] PackageCompiler: compiling incremental system imageinvalid redefinition of constant Package1
Stacktrace:
 [1] top-level scope
   @ none:1
⡆ [00m:08s] PackageCompiler: compiling incremental system imageModule, e::Any)
   @ Core ./boot.jl:373
 [3] top-level scope
   @ /tmp/jl_CY4U4d:15
 [4] eval(m::Module, e::Any)
   @ Core ./boot.jl:373
 [5] top-level scope
   @ /tmp/jl_CY4U4d:10
in expression starting at /tmp/jl_CY4U4d:10
```

## Solution

I managed to get past this by skipping that part of code (effectively turning off precompilation statements, but I wasn't using them so no problem)
The image built without any issues and I could load it and have access to the both uuid versions of PackageC 

So I think the temporary solution could be to add a "disable precompile files" flag or just skip the code if no precompile files/statements were provided

Precompiling in this setup is problematic, so not sure how we could address this.
We would need to somehow have separate `PrecompileStagingArea`s for precompilation and figure out which statement belongs to which precompile group

staticfloat:
I think this is going to require the precompile statements to emit UUID-specific lines.  E.g. we need to add a `Base.get_module(UUID(...))` and then if we load two packages with the same name, instead of emitting something like `precompile(Tuple{typeof(Foo.fooifier), Int64})` we'd instead emit something like `precompile(Tuple{typeof(Base.get_module(UUID(...)).fooifier), Int64})`.

PR #794, opened 2023-03-16
Precompile/import into main flags & warnings for package name conflicts

krynju:
Changelog:
1. added `import_into_main` flag (default true), which allows you to control whether the top level packages will be imported into Main in the sysimage
2. added `disable_precompile` flag (default false), which allows you to turn off the precompile part of the build completely
3. added skip & warnings on the precompile name conflict https://github.com/JuliaLang/PackageCompiler.jl/issues/767
4. added skip & warnings on the import to Main name conflict https://github.com/JuliaLang/PackageCompiler.jl/issues/768

While this doesn't fix the #767 and #768 issues it makes them not error, which in most cases will be fine.
If not then the warning should point the user at some possible solution.
The current solutions to both of these problems are either using the flags added in this PR or modifying the precompile statements.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -309,7 +309,9 @@ function create_sysimg_object_file(object_file::String,
                             script::Union{Nothing, String},
                             sysimage_build_args::Cmd,
                             extra_precompiles::String,
-                            incremental::Bool)
+                            incremental::Bool,
+                            import_into_main::Bool,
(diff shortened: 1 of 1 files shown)


staticfloat:
FWIW, using `import_into_main = false` is sufficient for me to build [this example case](https://github.com/JuliaLang/PackageCompiler.jl/issues/768#issuecomment-1487602317).

PR #802, opened 2023-04-06
split up the julia_init and sysimage into two libraries in `create_library`

KristofferC:
This avoids having to pass so many "internal" arguments from `create_library` to the sysimage builder itself.


KristofferC:
This doesn't work.

PR #806, opened 2023-05-07
Parallel parse / eval / precompile statements

ktdq:


ktdq:
Sorry for the sloppy code, but it does speed up sysimage creation, especially on nightly. Could probably be improved further.

ktdq:
Will resubmit cleaner

Issue #725, opened 2022-09-22
too much stuff generated by default


JeffBezanson:
If I generate a simple app, lib/julia contains
```
-rw-rw-r-- 1 jeff jeff  17734957 Sep 20 16:07 corecompiler.ji
-rw-rw-r-- 1 jeff jeff 234258408 Sep 20 16:07 sys-debug-o.a
-rwxrwxr-x 1 jeff jeff 201407816 Sep 20 16:07 sys-debug.so
-rw-rw-r-- 1 jeff jeff 106505394 Sep 20 16:07 sys.ji
-rw-rw-r-- 1 jeff jeff 211870856 Sep 20 16:07 sys-o.a
-rwxrwxr-x 1 jeff jeff  72735752 Sep 20 16:18 sys.so
```
Now, many rightly complain that the 70MB sys.so is too big, but this is actually not too bad by our standards! However this is somewhat undermined by the presence of *730MB* of extra stuff that is not needed to run the app. The other files should not be included by default. The .ji and .a files can always be deleted, and the debug image should be optional. Other low-hanging fruit would be to delete .a files in lib/ as well, which removes another ~180MB.

KristofferC:
PackageCompiler is not very smart when you use a local build that includes a bunch of extra stuff that is not present in a release build that your download. But we can teach it, yeah

xgdgsc:
I get a 300MB sys.so , is it expected?

Octogonapus:
Was this indeed fixed by #823, and should be closed?

PR #823, opened 2023-06-07
only bundle libraries that are needed

KristofferC:


KristofferC:
I think this should be ready to go.

sloede:
Out of curiosity, what will this PR being to PC.jl? From the title I assume it will improve somehow which libraries are bundled with a sysimage, but I can't tell from looking at the code 😬

KristofferC:
It wil:

- Only bundle libraries for jlls that are actually used in the resulting sysimage
- It will fix https://github.com/JuliaLang/PackageCompiler.jl/issues/725.

sloede:
> It will:
> 
> * Only bundle libraries for jlls that are actually used in the resulting sysimage
> * It will fix [too much stuff generated by default #725](https://github.com/JuliaLang/PackageCompiler.jl/issues/725).

Great, thanks!

PR #848, opened 2023-09-16
Allow passing projects (and not just packages) to `create_library`

### LINKED ISSUE #840: Preferences not copied into project that is compiled
When building a package that reads in preferences, PackageCompiler should write out the unified set of preferences that the to-be-compiled project can see, [just as `Pkg` does in its `sandbox` function](https://github.com/JuliaLang/Pkg.jl/blob/f570abd39d0a6ab483e658cf65c69af59a152314/src/Operations.jl#L1786-L1791).  As an example, I have created [this repository](https://github.com/staticfloat/package-compiler-preferences-test) which showcases how compile-time preferences (and runtime preferences) are not picked up as I would expect.  Just clone the repository and run `make`, it will run the `test.jl` file, which has comments showing each test and what it expects to get from each.

I believe compile-time preferences are a bug in PackageCompiler (it should copy the preferences over into its internal project/load path during package compilation) and runtime preferences not changing are just due to me not understanding how the environment stack works inside of the PackageCompiler regime.

sloede:
This is an attempt to resolve #840, using the suggestion by @staticfloat in https://github.com/JuliaLang/PackageCompiler.jl/issues/840#issuecomment-1721601096.

Before, `create_library` only accepted a `package_dir` to a directory with a `Project.toml` that describes an actual package, i.e., with a `name` and `uuid`. This PR relaxes this requirement and allows one to also pass a mere project directory to `create_library`.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1068,24 +1068,6 @@ function create_sysimage_workaround(
                     soname::Union{Nothing,String},
                     script::Union{Nothing,String}
                     )
-    # If environment is not a package but a project, directly create sysimage
-    if ctx.env.pkg === nothing
-        project = dirname(ctx.env.project_file)
-        create_sysimage(String[]; sysimage_path, project,
-                        incremental,
-                        script,
-                        precompile_execution_file,
-                        precompile_statements_file,
-                        cpu_target,
-                        julia_init_c_file,
-                        version,
-                        soname,
-                        sysimage_build_args,
-                        include_transitive_dependencies)
-        return
-    end
-
-    package_name = ctx.env.pkg.name
     project = dirname(ctx.env.project_file)
 
     if !incremental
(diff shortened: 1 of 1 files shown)


sloede:
At the moment, this PR is not functional, since the `@ccallable` functions from my dependencies do not end up in the final library. Do you see where I went wrong here @KristofferC, seeing that it was you who implemented a lot of the surrounding functionality that I touched?

sloede:
OK, I seem to have found the issue and fixed it. From some initial tests, this seems to work. It would be great to get some feedback from @staticfloat or @KristofferC regarding whether this is on the right track to being mergeable.

staticfloat (approved):
Seems reasonable to me, but @KristofferC is the expert here.

KristofferC:
Nightly CI should be fixed so rerunning it.

PR #572, opened 2021-10-02
bundle Julia executable and Project.toml to get Distributed working

KristofferC:
Fixes https://github.com/JuliaLang/PackageCompiler.jl/issues/84 and fixes https://github.com/JuliaLang/PackageCompiler.jl/issues/161.

PR #850, opened 2023-09-20
Bundle `Project.toml` with libraries

sloede:
Similar to #572, this PR bundles the `Project.toml` file with libraries. The motivation is that otherwise it is not possible to run dynamic code in the library that is `using` one of the project dependencies. For example, when `include`ing a file from a library function, `using MyPkg` will currently not work even if `MyPkg` is listed in the `[deps]` section of the project dir. With this PR, `using MyPkg` will be possible again.

I also added a note to the docstring explaining that installing multiple apps/libraries to the same prefix is generally not possible, since they will overwrite each other's `Project.toml` in `<prefix>/share/julia`. This was already the case before for apps, I just amended the documentation for it.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -714,6 +714,9 @@ end
 
 The executable will be placed in a folder called `bin` in `compiled_app` and
 when the executable run the `julia_main` function is called.
+Note that since an app-specific `Project.toml` is placed in the `share/julia` folder in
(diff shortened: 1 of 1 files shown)


sloede:
@KristofferC @IanButterworth this is ready for review from my point of view.

sloede:
Thanks a lot for the fast response! It would be great if you could also tag a patch release.

PR #852, opened 2023-09-21
Bundle preferences with apps/libraries

sloede:
This PR attempts to address those parts of #840 that were not covered by #848. It adds a new function `bundle_preferences` that will create a `LocalPreferences.toml` file with all preferences that can be seen by the package or project for which an app/library is compiled. The preferences will will be included in the app/library bundle under `PREFIX/share/julia` (which is the default `LOAD_PATH`), such that they get picked up automatically.

@staticfloat Do you think that this would adequately address your original issue in #840?

### DIFF
--- a/docs/src/apps.md
+++ b/docs/src/apps.md
@@ -202,6 +202,18 @@ that does some simple arithmetic. It is instructive to see how the [artifact
 file](https://github.com/JuliaLang/PackageCompiler.jl/blob/master/examples/MyApp/Artifacts.toml)
 is [used in the source](https://github.com/JuliaLang/PackageCompiler.jl/blob/d722a3d91abe328ebd239e2f45660be35263ebe1/examples/MyApp/src/MyApp.jl#L7-L8).
 
+### [Preferences](@id app-preferences)
+
+Compile-time preferences used by any of the packages included in the app will be stored in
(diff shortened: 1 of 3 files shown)


sloede:
I see that `Base.get_preferences()` was not available in Julia v1.6. Is there an alternative way to achieve the same thing, i.e., to get a Dict with all preferences and their values? Or would I have to copy the relevant parts from https://github.com/JuliaLang/julia/blob/master/base/loading.jl to make it work for older Julia versions?

staticfloat:
> I see that Base.get_preferences() was not available in Julia v1.6.

You.... could.  I think preference loading and whatnot should probably work on older Julia versions, the main piece of functionality that is missing is the invalidation of compile-time preferences, but that's not an issue in this case as you're compiling only once.  In general though, we don't usually encourage people to use preferences on Julia versions earlier than v1.6, as it can get a little confusing when things don't work the same across all versions.

sloede:
@KristofferC I've been using this PR successfully for the past week. Thus it would be great to get a review 🙂

sloede:
@KristofferC @staticfloat if there are no objections, I will merge this within the next few days.

staticfloat (approved):

PR #853, opened 2023-09-27
Bundle dynamically selected libstdc++

sloede:
Julia by default uses [`libstdcxxprobe()`](https://github.com/JuliaLang/julia/blob/8de80bdf902e0899dd3e39636a95769cf44c823b/cli/loader_lib.c#L254) on Linux systems to figure out whether to load the system `libstdc++` or the one bundled with Julia. Not doing this can cause problems if other libraries used by a Julia process are compiled against a system libstdc++ that is newer than the one provided by Julia

PackageCompiler currently always bundles the library shipped with Julia, not honoring the dynamic selection. This PR aims to remedy this by searching for the `libstdc++` that will be bundled in the directory where the currently loaded `libstdc++` resides.

xref: https://discourse.julialang.org/t/precedence-of-local-and-julia-shipped-shared-libraries/104258?u=sloede

@staticfloat it would be great if you could have a look if this seems like a reasonable implementation to you.

@giordano thanks a lot for the initial pointer on how get the `libstdc++` actually used by Julia.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1183,66 +1183,52 @@ function glob_pattern_lib(lib)
     error("unknown os")
 end
 
-# Take `path` to a libstdc++ library and return aliases for soname symlinks
+# Take `path` to a libstdc++ library and return aliases for shared library symlinks:
 # - Unix:
 #   `path/to/libstdc++.so.6.0.30` will yield aliases
 #   `libstdc++.so.6.0.30`, `libstdc++.so.6`, `libstdc++.so`
 # - Apple:
 #   `path/to/libstdc++.6.0.30.dylib` will yield aliases
 #   `libstdc++.6.0.30.dylib`, `libstdc++.6.dylib`, `libstdc++.dylib`
-# - Windows (no versions in filenames):
-#   `path/to/libstdc++.dll` will yield alias `libstdc++.dll`
+# - Windows:
+#   `path/to/libstdc++-6.0.30.dll` will yield aliases
+#   `libstdc++-6.0.30.dll`, `libstdc++-6.dll`, `libstdc++.dll`
+# If the major version cannot be inferred from the filename, return only full name and the
+# alias without any version information.
 function get_libstdcxx_aliases(path)
-    # Always add full filename to list of aliases
-    filename = basename(path)
-    aliases = String[filename]
-
-    if Sys.iswindows()
-        # Windows does not encode versions in filenames
-        return aliases
-    end
-
-    # Split filename into components
-    components = split(filename, ".")
-
-    if Sys.isapple()
-        # macOS uses `libname[.major[.minor[.patch]].dylib`, where the version parts are optional
-        # we want the full name (`filename`) plus
-        # - `libname.dylib`
-        # - `libname.major.dylib` (if `major` exists)
-        if last(components) != Libdl.dlext
-            error("unexpected filename: last component should be `.$(Libdl.dlext)` but is `$(last(components))`")
-        end
-        libname = components[1]
-        ext = components[end]
-        push!(aliases, libname * "." * ext) # `libname.dylib`
-        if length(components) > 2
-            major = components[2]
-            push!(aliases, libname * "." * major * "." * ext) # `libname.major.dylib`
-        end
-        return unique(aliases)
+    # Extract library name and version from path
+    os = if Sys.iswindows()
+        "windows"
+    elseif Sys.isapple()
+        "macos"
     elseif Sys.isunix()
-        # Unix uses `libname.so[.major[.minor[.patch]]`, where the version parts are optional
-        # we want the full name (`filename`) plus
-        # - `libname.so`
-        # - `libname.major.so` (if `major` exists)
-        if length(components) < 2
-            error("unexpected filename: no file extension found")
-        end
-        if components[2] != Libdl.dlext
-            error("unexpected filename: second component should be `.$(Libdl.dlext)` but is `$(components[2])`")
-        end
-        libname = components[1]
-        ext = components[2]
-        push!(aliases, libname * "." * ext) # `libname.so`
-        if length(components) > 2
-            major = components[3]
-            push!(aliases, libname * "." * ext * "." * major) # `libname.so.major`
-        end
-        return unique(aliases)
+        "unix"
     else
         error("unable to determine aliases; system is not Windows, macOS, or UNIX")
     end
+    libname, version = Base.BinaryPlatforms.parse_dl_name_version(path, os)
+
+    # Always add full filename and base library name to list of aliases
+    ext = Libdl.dlext
+    aliases = String[basename(path), "$libname.$ext"]
+
+    # If version exists, also add alias for major version
+    if !isnothing(version)
+        major = version.major
+        if Sys.iswindows()
+            # libname-major.dll
+            push!(aliases, "$libname-$major.$ext")
+        elseif Sys.isapple()
+            # libname.major.dylib
+            push!(aliases, "$libname.$major.$ext")
+        else
+            # libname.so.major
+            push!(aliases, "$libname.$ext.$major")
+        end
+    end
+
+    # Return unique list in case, e.g., filename is identical to base library name
+    return unique(aliases)
 end
 
 # TODO: Detangle printing from business logic
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1200,11 +1200,17 @@ function bundle_julia_libraries(dest_dir, stdlibs)
     tot_libsize = 0
     printstyled("PackageCompiler: bundled libraries:\n")
 
-    # Reqiored libraries
+    # Bundle the libstdc++ that is actually loaded by Julia
+    # xref: https://discourse.julialang.org/t/precedence-of-local-and-julia-shipped-shared-libraries/104258?u=sloede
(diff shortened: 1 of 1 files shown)


src/PackageCompiler.jl
```diff
@@ -1200,11 +1200,23 @@ function bundle_julia_libraries(dest_dir, stdlibs)
...
+    # Bundle the libstdc++ that is actually loaded by Julia
+    # xref: https://discourse.julialang.org/t/precedence-of-local-and-julia-shipped-shared-libraries/104258?u=sloede
+    libstdcpp_search = filter(contains("libstdc++"), Libdl.dllist())
+    if isempty(libstdcpp_search)
+        # Fall back to default search directory if libstdc++ is not loaded
+        libstdcpp_dir = libjulia_dir
+    else
+        # If libstdc++ was found, get absolute path to directory
+        libstdcpp_dir = dirname(abspath(first(libstdcpp_search)))
+    end
```
staticfloat, line 1212: Another option is to just `dlopen("libstdc++")` before running `Libdl.dllist()`:

On macOS that will load the one that comes with Julia:
```
julia> using Libdl
       Libdl.dllist()
       @show filter(contains("libstdc++"), Libdl.dllist())
       dlopen("libstdc++")
       @show filter(contains("libstdc++"), Libdl.dllist());
filter(contains("libstdc++"), Libdl.dllist()) = String[]
filter(contains("libstdc++"), Libdl.dllist()) = ["/Users/sabae/.julia/juliaup/julia-1.9.3+0.aarch64.apple.darwin14/lib/julia/libstdc++.6.dylib"]
```

On Linux (where we do the fancy `libstdc++` search) you'll already have `libstdc++` loaded, so the `dlopen("libstdc++")` doesn't do anything.  The only reason I suggest this is that then `Libdl.dllist()` will always then give you the absolute path to the file, which may be less prone to edge cases than searchinmg in a directory.

sloede: You think it could happen that `Libdl.dllist()` contains a library without a full path?

The reason I implemented it similar to the existing search is that this way you end up with, e.g.,
```
lrwxrwxrwx 1 mschlott mi      19 Sep 27 09:17 libstdc++.so.6 -> libstdc++.so.6.0.28
-rw-r--r-- 1 mschlott mi 1956992 Sep 27 09:17 libstdc++.so.6.0.28
```
in `PREFIX/lib/julia` for a `/usr/lib/x86_64-linux-gnu` directory with
```
lrwxrwxrwx 1 root root      19 Jul  9 08:45 /usr/lib/x86_64-linux-gnu/libstdc++.so.6 -> libstdc++.so.6.0.28
-rw-r--r-- 1 root root 1956992 Jul  9 08:45 /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28
```
Otherwise, you'd have to `cp(src, dst; follow_symlinks=true)` to make sure that symbolic links are resolved before copying.

But if you think that's better/saner/more robust, I'm happy to change it as you suggested.

staticfloat: The reason for `libstdc++.so.6 -> libstdc++.so.6.0.28` to exist is so that a wide variety of applications can link against `libstdc++.so.6` and find the file on-disk, but once it's already in memory that shouldn't matter, it should look at the `SONAME` of the loaded library and short-circuit based on that.  I like using `Libdl.dllist()`'s output (which should always be an absolute pathname to the actual file) to avoid problems like finding a linker script instead of the actual `.so`, bundling multiple versions of the same library if you somehow have multiple, etc....  If you need a library, I just think it's simpler to copy precisely the file that ends up getting loaded.

sloede: I tried your suggestion, but I ran into new troubles. On my test system, the system libstdc++ is actually older than the one bundled with Julia. Thus, before I used
```shell
LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28 my/packagecompiler/build/command
```
to make sure that I can test with using a non-Julia libstdc++. In this case, however, Julia will actually load *both* libraries, and
```julia
dlopen("libstdc++")
@show filter(contains("libstdc++"), Libdl.dllist());
```
yields
```
["/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28", "/mnt/hd1/opt/julia/1.9.3/bin/../lib/julia/libstdc++.so"]
```

Thus it seems if someone is intent on preloading libstdc++ (which is not unheard of), this new approach will fail - indeed, `dlopen` *will* load a second library if they do not match 100%.

I then tried the following, where I tested whether libstdc++ is already loaded and then call `Libdl.dlopen` only if it isn't. This remedies the problem before, but now a second issue arises:

While `dllist()` contains absolute paths, it does _not_ resolve symbolic links. Since Julia looks for `libstdc++.so.6`, it will stop and use the first it finds, e.g., `/mnt/hd1/opt/julia/1.9.3/bin/../lib/julia/libstdc++.so.6`. However, I cannot just copy this path, since it is just a symbolic link. If I call `realpath` before copying, this will, however, put `libstdc++.so.6.0.30` in my bundle, which will _not_ be picked up when Julia searches for `libstdc++.so.6` (I verified this using `LD_DEBUG=libs`). Thus, the next option would be that I copy the library _without_ resolving symlinks, but use `follow_symlinks=true`. This will work well, but now we end up with only `libstdc++.so.6` in the bundle without being able to tell which version it actually is. It also seems fishy to me to leave it at that, given that is _not_ how the symlinks are organized on Linux/macOS systems.

Therefore, what I seem to end up with (as far as I can tell), is the following:
* Use `dlopen` to load `libstdc++` if not yet present in `dllist()`
* Get path to `libstdc++` from `dllist()`
* Expand path to `realpath(abspath(...))`
* Copy this `libstdc++` to `PREFIX/lib/julia` (e.g., `libstdc++.so.6.0.28` or `libstdc++.6.dylib` or `libstdc++.dll`)
* Create relative symlinks within `PREFIX/lib/julia`:
  * `Sys.iswindows()`: nothing to do
  * `Sys.isapple()`:
    ```
    libstdc++.dylib -> libstdc++.6.dylib
    ```
  * `Sys.isunix()`:
    ```
    libstdc++.so -> libstdc++.so.6.0.28
    libstdc++.so.6 -> libstdc++.so.6.0.28
    ```

This seems to be very involved 😅, thus it would be great to get some feedback whether this sounds correct or if I'm missing something else.

sloede: I went ahead and implemented this newest approach in https://github.com/JuliaLang/PackageCompiler.jl/pull/853/commits/21c2f726ccd78fe95f35df2d954b864bf8020330. I am having it tested it by the colleague who filed the original issue, and we will report back.

bgeihe: Hi there!

I just confirmed that  @sloede 's current implementation fixes my initial issue, i.e. with his fix my system `libstdc++.so.6.0.32` is correctly picked up by PackageCompiler.

sloede:
@staticfloat it would be great to get another review from you, otherwise I'll go ahead and merge this within the next few days (unless @KristofferC objects).

src/PackageCompiler.jl
```diff
@@ -1183,6 +1183,68 @@ function glob_pattern_lib(lib)
...
+# Take `path` to a libstdc++ library and return aliases for soname symlinks
+# - Unix:
+#   `path/to/libstdc++.so.6.0.30` will yield aliases
+#   `libstdc++.so.6.0.30`, `libstdc++.so.6`, `libstdc++.so`
+# - Apple:
+#   `path/to/libstdc++.6.0.30.dylib` will yield aliases
+#   `libstdc++.6.0.30.dylib`, `libstdc++.6.dylib`, `libstdc++.dylib`
+# - Windows (no versions in filenames):
+#   `path/to/libstdc++.dll` will yield alias `libstdc++.dll`
+function get_libstdcxx_aliases(path)
```
staticfloat, line 1195: If you like, you can make use of `Base.BinaryPlatforms.parse_dl_name_version` to eliminate some of the toil here.

staticfloat (approved):
Overall, this looks like a fine approach to me!

src/PackageCompiler.jl
sloede: Thanks for the hint! Done in https://github.com/JuliaLang/PackageCompiler.jl/pull/853/commits/8fbe560a534d58a4a0ae3d8e191e383ab1eee1ab.

sloede:
@bgeihe could you please try once more if this PR works on your previously broken system, now that I simplified the library alias code? If it does (and the tests pass), we can merge this PR.

bgeihe:
Confirmed!

PallHaraldsson:
It good that you fixed this, but I see what we bundle has grow from 1378688 libstdc++.so.6.0.25, to 18006768 julia-1.8.1/usr/lib/libstdc++.so.6.0.29 to 18763496 libstdc++.so.6.0.30 and is non-small already.

So is 13% larger, and is the 18 MB overhead actually really needed? We're updating to LLVM 16, should likely 17 and 18, and can we expect we need to update newer larger each time?

I was thinking couldn't Julia apps in most cases just dynamically probe and use the system C++ library, at least at some point in the future? I'm not sure what other programs do, C++ based or otherwise.

I believe the C++ library is used by LLVM, not the rest of Julia, and LLVM can optionally be dropped. Would it then be ok to drop that library too?

I realize some JLL (e.g. for Qt) might also use it, but how recent C++ system library can we expect? Locate actually located none for me on my Linux Mint, which is strange, you would always have at least some version?!

PR #870, opened 2023-10-20
[WIP] Allow setting a command prefix for the sysimage build command

sloede:
This PR adds a new, optional `sysimage_build_prefix` kwarg to `create_{sysimage,app,library}`. It aims to allow one to prepend the build command called when creating a sysimage with a user-specified command prefix.

The main motivation is to be able to call PackageCompiler.jl from within an MPI-started process. This is sometimes required on supercomputers, where calling MPI functions without having started a process via `mpirun`/`srun`/`aprun` etc. is disabled.

cc @lchristm


lchristm:
@sloede The prefix is only needed for executing scripts passed via `precompile_execution_file` and not for building the system image afterwards. I changed the PR branch accordingly but I think it would be reasonable to change the PR title as well.

PR #877, opened 2023-10-25
Bundle `7z[.exe]` with libaries/apps

### LINKED ISSUE #681: `7z.exe` and `7z.dll` are not included as dependencies on the bin folder of a compiled package.
I am distributing some compiled versions of packages and recently a user reported that the module broke because he had no permission to check if there was a 7z library when using the command `Sys.which(p7zip_exe)` in p7zip_jll. This made me realize that 7z was not being bundled together with other libs since it didn't return after checking if the file existed in the following function.
```julia
function init_p7zip_path()
    # Prefer our own bundled p7zip, but if we don't have one, pick it up off of the PATH
    # If this is an in-tree build, `7z` will live in `bin`.  Otherwise, it'll be in `libexec`
    for bundled_p7zip_path in (joinpath(Sys.BINDIR, Base.LIBEXECDIR, p7zip_exe),
                               joinpath(Sys.BINDIR, p7zip_exe))
        if isfile(bundled_p7zip_path)
            global p7zip_path = abspath(bundled_p7zip_path)
            return
        end
    end
    global p7zip_path = something(Sys.which(p7zip_exe), p7zip_exe)
end
```
It happens that `7z.dll` and `7z.exe` are not present in the `bin` folder of julia but are in `Base.LIBEXECDIR`. Apparently when we compile a package all the files in `Base.LIBEXECDIR` are not copied together with other dependencies. 

Could we add a feature that allows users to add `7z` to the libs along with everyother dependency?

sloede:
Resolves #681.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1366,7 +1366,7 @@ function bundle_julia_libraries(dest_dir, stdlibs)
     return
 end
 
-function bundle_julia_libexec(ctx, destdir)
+function bundle_julia_libexec(ctx, dest_dir)
     # We only bundle the `7z` executable at the moment
     @assert ctx.env.manifest !== nothing
     uses_p7zip_jll = false
@@ -1380,8 +1380,14 @@ function bundle_julia_libexec(ctx, destdir)
         return
     end
 
-    # Use Julia-private `libexec` folder (normpath is required in case `bin` does not exist)
-    bundle_libexec_dir = normpath(joinpath(dest_dir, "bin", Base.PRIVATE_LIBEXECDIR))
+    # Use Julia-private `libexec` folder if it exsts
+    # (normpath is required in case `bin` does not exist in `dest_dir`)
+    libexecdir_rel = if isdefined(Base, :PRIVATE_LIBEXECDIR)
+        Base.PRIVATE_LIBEXECDIR
+    else
+        Base.LIBEXECDIR
+    end
+    bundle_libexec_dir = normpath(joinpath(dest_dir, "bin", libexecdir_rel))
     mkpath(bundle_libexec_dir)
 
     p7zip_exe = basename(p7zip_path)
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1369,14 +1369,7 @@ end
 function bundle_julia_libexec(ctx, dest_dir)
     # We only bundle the `7z` executable at the moment
     @assert ctx.env.manifest !== nothing
-    uses_p7zip_jll = false
-    for (_, package) in ctx.env.manifest
-        if package.name == "p7zip_jll"
-            uses_p7zip_jll = true
-            break
-        end
-    end
-    if !uses_p7zip_jll
+    if !any(x -> x.name == "p7zip_jll", values(ctx.env.manifest))
         return
     end
 
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1039,7 +1039,7 @@ function create_library(package_or_project::String,
     mkpath(dest_dir)
     stdlibs = filter_stdlibs ? gather_stdlibs_project(ctx; only_in_sysimage=false) : _STDLIBS
     bundle_julia_libraries(dest_dir, stdlibs)
-    bundle_julia_libexec(ctx, destdir)
+    bundle_julia_libexec(ctx, dest_dir)
     bundle_artifacts(ctx, dest_dir; include_lazy_artifacts)
     bundle_headers(dest_dir, header_files)
     bundle_project(ctx, dest_dir)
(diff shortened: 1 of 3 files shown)


staticfloat:
Do we need to bundle `lld` as well?

sloede:
> Do we need to bundle `lld` as well?

Didn't see any issues complaining about missing `lld`, while `7z` is required by Pkg, and it's been complained about just today by two separate parties. I thus went ahead and went with the minimum solution that can help to fix this issue.

Though, as discussed on Slack today, it does not fix the underlying issue of, e.g., MKL_jll being downloaded even if we do not want it.

src/PackageCompiler.jl
```diff
@@ -1363,6 +1365,38 @@ function bundle_julia_libraries(dest_dir, stdlibs)
...
+    has_pkg = false
+    for (_, package) in ctx.env.manifest
+        if package.name == "Pkg"
+            has_pkg = true
+            break
+        end
+    end
+    if !has_pkg
+        return
+    end
```
staticfloat, line 1380: I think a better check would be if `p7zip_jll` is in the manifest, [which Pkg depends on](https://github.com/JuliaLang/Pkg.jl/blob/487f9e271d7ed48eb336409879e793149977d5a0/Project.toml#L25).

src/PackageCompiler.jl
```diff
@@ -1363,6 +1365,38 @@ function bundle_julia_libraries(dest_dir, stdlibs)
...
+
+    # We only bundle the `7z` from Julia
+    for julia_p7zip_path in (joinpath(Sys.BINDIR, Base.PRIVATE_LIBEXECDIR, p7zip_exe),
+                             joinpath(Sys.BINDIR, p7zip_exe))
+        if isfile(julia_p7zip_path)
+            mkdir(bundle_libexec_dir)
+            cp(julia_p7zip_path, bundle_p7zip_path)
+            return
+        end
+    end
```
staticfloat, line 1394: Perhaps it's just easier to use `p7zip_jll.p7zip_path`?

staticfloat (changes requested):

src/PackageCompiler.jl
sloede: Sounds good!

src/PackageCompiler.jl
sloede: I thought about this. However, in trying to remain at least somewhat portable, I decided against it and figured it might be better to just bundle whatever Julia is shipping. Otherwise, we might bundle a `7z` executable that is not usable on other machines due to some secondary dependencies that are only available on the host system. But if you think this is a non-concern, I could also use `p7zip_jll.p7zip_path`

staticfloat: > I decided against it and figured it might be better to just bundle whatever Julia is shipping

`p7zip_jll.p7zip_path` _is_ what Julia is shipping. :)  JLLs that ship with Julia give you the path to the thing that ships with Julia.

sloede: From what I understand [here](https://github.com/JuliaLang/julia/blob/d0c42840f6ff509421a3f0d28ebe5bd80c625cc3/stdlib/p7zip_jll/src/p7zip_jll.jl#L70-L81), it could also be the executable that is found through `Sys.which`. Or am I missing something?

staticfloat: That is correct; but I don't expect such a Julia to be a common occurrence, and in the future if our stdlib JLLs become more JLL-like (and no longer dump things into `${prefix}/libexec` but rather into `${prefix}/share/julia/stdlib/v1.X/artifacts/1a2b3c4d/`) that path will continue to be correct, whereas the one here will no longer be correct.

sloede: OK, gotcha. But that means I should also make `p7zip_jll` explicitly a dependency of PackageCompiler.jl, right?

staticfloat: Yes

src/PackageCompiler.jl
sloede: Done in c87edb8.

src/PackageCompiler.jl
sloede: Done in c87edb8.

sloede:
@bgeihe @mashu it would be great if you could test this PR to verify that it fixes your respective issues with `7z` not being found by Pkg.

src/PackageCompiler.jl
```diff
@@ -1363,6 +1366,36 @@ function bundle_julia_libraries(dest_dir, stdlibs)
     return
 end
 
+function bundle_julia_libexec(ctx, dest_dir)
+    # We only bundle the `7z` executable at the moment
+    @assert ctx.env.manifest !== nothing
+    uses_p7zip_jll = false
+    for (_, package) in ctx.env.manifest
```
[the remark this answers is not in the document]
sloede: Thanks for the suggestion! Done in 153c30f.

bgeihe:
Works for me. Praise @sloede !

staticfloat (approved):

bjarthur:
i tested this PR and it works great!  fixes https://github.com/JuliaLinearAlgebra/MKL.jl/issues/137 too

sloede:
@bjarthur Thanks a lot for verifying! We have conflicting reports from another use (@bgeihe) so I would like to run a few more tests before merging this. However, since for the majority of people this PR seems to work fine, I will create a new release with it probably still within this week.

PR #878, opened 2023-10-26
Update LOAD_PATH in `dump_preferences` to include `@stdlibs`

### LINKED ISSUE #873: Required TOML.jl
Hi,

v2.1.10 seems to require TOML.jl for the target package (that is built from), see error message.

I am using `create_library` with the following keywords (might be an unusual configuration):
```
PackageCompiler.create_library(...;
                                    lib_name=...,
                                    precompile_execution_file=...,
                                    precompile_statements_file=...,
                                    incremental=false,
                                    filter_stdlibs=false,
                                    julia_init_c_file = ...,
                                    header_files = ..., 
                                    force=true,
                                    include_transitive_dependencies=true,
                                    include_lazy_artifacts=true)
```
----------------------------
Error:
```
ERROR: ArgumentError: Package TOML not found in current path.
- Run `import Pkg; Pkg.add("TOML")` to install the TOML package.
Stacktrace:
 [1] macro expansion
   @ .\loading.jl:1630 [inlined]
 [2] macro expansion
   @ .\lock.jl:267 [inlined]
 [3] require(into::Module, mod::Symbol)
   @ Base .\loading.jl:1611

  LoadError: failed process: Process(`'C:\Users\...\AppData\Local\Programs\Julia-1.9.3\bin\julia.exe' -Cnative '-JC:\Users\...\AppData\Local\Programs\Julia-1.9.3\lib\julia\sys.dll' --depwarn=yes --check-bounds=yes -g1 --color=yes --startup-file=no '--project=C:\Users\...\AppData\Local\Temp\...' -e 'using TOML, Pkg
  # For each dependency pair (UUID => PackageInfo), store preferences in Dict
  prefs = Dict{String,Any}(last(dep).name => Base.get_preferences(first(dep)) for dep in Pkg.dependencies())
  # Filter out packages without preferences
  filter!(p -> !isempty(last(p)), prefs)
  TOML.print(prefs, sorted=true)
  '`, ProcessExited(1)) [1]

  Stacktrace:
    [1] pipeline_error
      @ .\process.jl:565 [inlined]
    [2] read(cmd::Cmd)
      @ Base .\process.jl:449
    [3] read
      @ .\process.jl:458 [inlined]
    [4] dump_preferences(io::IOStream, project_dir::String)
      @ PackageCompiler C:\Users\...\.julia\packages\PackageCompiler\LLlSZ\src\PackageCompiler.jl:1536
    [5] #39
      @ C:\Users\...\.julia\packages\PackageCompiler\LLlSZ\src\PackageCompiler.jl:1552 [inlined]
    [6] open(::PackageCompiler.var"#39#40"{String}, ::String, ::Vararg{String}; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
      @ Base .\io.jl:395
    [7] open
      @ .\io.jl:392 [inlined]
    [8] bundle_preferences(ctx::Pkg.Types.Context, dest_dir::String)
      @ PackageCompiler C:\Users\...\.julia\packages\PackageCompiler\LLlSZ\src\PackageCompiler.jl:1551
    [9] create_library(package_or_project::String, dest_dir::String; lib_name::String, precompile_execution_file::Vector{String}, precompile_statements_file::Vector{String}, incremental::Bool, filter_stdlibs::Bool, force::Bool, header_files::Vector{String}, julia_init_c_file::String, julia_init_h_file::String, version::Nothing, compat_level::String, cpu_target::String, include_lazy_artifacts::Bool, sysimage_build_args::Cmd, include_transitive_dependencies::Bool, include_preferences::Bool, script::Nothing)
      @ PackageCompiler C:\Users\...\.julia\packages\PackageCompiler\LLlSZ\src\PackageCompiler.jl:1044
```

Message is indeed related to [this line](https://github.com/JuliaLang/PackageCompiler.jl/blob/846292c50a7177ea27f1fdb607be8c42174e030e/src/PackageCompiler.jl#L1531).

In Julia v1.9.3 with PackageCompiler v2.1.10 only - v2.1.9 is working without issues.

Thanks in advance,
ThummeTo

sloede:
Hopefully fixes #873.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1535,27 +1535,17 @@ function dump_preferences(io::IO, project_dir)
     # Note: in `command` we cannot just use `Base.get_preferences()`, since this API was
     #       only introduced in Julia v1.8
     command = """
+    # Ensure that `@stdlib` is part of `LOAD_PATH` such that we get TOML and Pkg
+    pushfirst!(LOAD_PATH, "@stdlib")
     using TOML, Pkg
+    popfirst!(LOAD_PATH)
     # For each dependency pair (UUID => PackageInfo), store preferences in Dict
     prefs = Dict{String,Any}(last(dep).name => Base.get_preferences(first(dep)) for dep in Pkg.dependencies())
     # Filter out packages without preferences
     filter!(p -> !isempty(last(p)), prefs)
     TOML.print(prefs, sorted=true)
     """
-
-    cmd = `$(Base.julia_cmd()) --project=$project_dir -e "$command"`
-
-    # Ensure that `@stdlib` is part of `LOAD_PATH` such that we get TOML and Pkg
-    if !("@stdlib" in LOAD_PATH)
-        splitter = Sys.iswindows() ? ';' : ':'
-        new_load_path = get(ENV, "JULIA_LOAD_PATH", "") * "$(splitter)@stdlib"
-        cmd = addenv(cmd, "JULIA_LOAD_PATH" => "$new_load_path")
-        @debug "dump_preferences: running $cmd" JULIA_LOAD_PATH = "$new_load_path"
-    else
-        @debug "dump_preferences: running $cmd"
-    end
-
-    prefs = read(cmd, String)
+    prefs = read(`$(Base.julia_cmd()) --project=$project_dir -e "$command"`, String)
     write(io, prefs)
 
     nothing
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1548,7 +1548,7 @@ function dump_preferences(io::IO, project_dir)
     # Ensure that `@stdlib` is part of `LOAD_PATH` such that we get TOML and Pkg
     if !("@stdlib" in LOAD_PATH)
         splitter = Sys.iswindows() ? ';' : ':'
-        new_load_path = join(vcat(LOAD_PATH, "@stdlib"), splitter)
+        new_load_path = get(ENV, "JULIA_LOAD_PATH", "") * "$(splitter)@stdlib"
         cmd = addenv(cmd, "JULIA_LOAD_PATH" => "$new_load_path")
         @debug "dump_preferences: running $cmd" JULIA_LOAD_PATH = "$new_load_path"
     else
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1542,7 +1542,20 @@ function dump_preferences(io::IO, project_dir)
     filter!(p -> !isempty(last(p)), prefs)
     TOML.print(prefs, sorted=true)
     """
-    prefs = read(`$(Base.julia_cmd()) --project=$project_dir -e "$command"`, String)
+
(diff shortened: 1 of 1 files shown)


sloede:
@fredrikekre it would be great if you could have a look if this is what you had in mind

fredrikekre:
Should be possible to always just use `@stdlib` as the complete load path. `--project` will set the active projet that Pkg will look at, but it isn't a problem that this isn't in the path, I think. The only thing I am not sure of is how preferences in other load path entries are handled then.

sloede:
> The only thing I am not sure of is how preferences in other load path entries are handled then.

Yeah, exactly. This is what has me worried as well, since the whole point of using an external process for getting the preferences is to make sure that we do not miss anything (or mix it up with another project).

Instead of using `LOAD_PATH` for it, I could also just set
```julia
new_load_path = get(ENV, "JULIA_LOAD_PATH", "") * "$(splitter)@stdlib"
```
That would prevent the `LOAD_PATH` as set for the current PC process to spill into the new process, but respect whatever the user has set in `JULIA_LOAD_PATH`.

fredrikekre:
To be completely idiot-proof I think you should always put `@stdlib` as the *first* entry, and then in the script, just after `using Pkg, TOML` pop this entry from the load path or something like that.

sloede:
> To be completely idiot-proof I think you should always put `@stdlib` as the _first_ entry, and then in the script, just after `using Pkg, TOML` pop this entry from the load path or something like that.

Are you sure that is necessary? In that case I would need to track if I added `@stdlib` manually, and if yes, remove it again. I feel like this is overkill.

But which situation did you have in mind that would be guarded against using this approach with `@stdlib` as the first entry and then removing it again?

fredrikekre:
> In that case I would need to track if I added @stdlib manually, and if yes, remove it again.

You should be able to do it unconditionally.

> But which situation did you have in mind [...]

A user could theoretically have another package named Pkg or TOML in their project (with different UUID). I also think the patch becomes smaller with this approach too:

```diff
diff --git a/src/PackageCompiler.jl b/src/PackageCompiler.jl
index 3624b35..572a7e1 100644
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1536,7 +1536,9 @@ function dump_preferences(io::IO, project_dir)
     # Note: in `command` we cannot just use `Base.get_preferences()`, since this API was
     #       only introduced in Julia v1.8
     command = """
+    pushfirst!(LOAD_PATH, "@stdlib")
     using TOML, Pkg
+    popfirst!(LOAD_PATH, "@stdlib")
     # For each dependency pair (UUID => PackageInfo), store preferences in Dict
     prefs = Dict{String,Any}(last(dep).name => Base.get_preferences(first(dep)) for dep in Pkg.dependencies())
     # Filter out packages without preferences
```

fredrikekre:
Please remove the @-mention from the squash later, looks like Github still pings when such commits are pushed...

fredrikekre (approved):

PR #883, opened 2023-10-30
Tweak how `IS_OFFICIAL` is determined (unbreaks Julia 1.11)

Drvi:
In https://github.com/JuliaLang/julia/pull/51867, the `banner` function moved to the `REPL` module, breaking precompilation.

Perhaps there is a better way to know whether we are on an official build, but we're trying to resolve this issue to unblock our nightly CI.


nickrobinson251 (approved):

src/PackageCompiler.jl
```diff
@@ -710,7 +711,11 @@ end
 # App #
 #######
 
-const IS_OFFICIAL = occursin("Official https://julialang.org/ release", sprint(Base.banner))
+const IS_OFFICIAL = occursin(
+    "Official https://julialang.org/ release",
+    # `banner` moved in https://github.com/JuliaLang/julia/pull/51867
+    sprint(isdefined(Base, :banner) ? Base.banner : REPL.banner)
+)
```
LilithHafner, line 718:
```suggestion
const IS_OFFICIAL = Base.GIT_VERSION_INFO.tagged_commit
```
This is the better way. Tested on 1.0.5, 1.9.3, and master.

This is how Julia decides whether or not to print that string
```julia
function banner(io::IO = stdout; short = false)
    if Base.GIT_VERSION_INFO.tagged_commit
        commit_string = Base.TAGGED_RELEASE_BANNER
    elseif isempty(Base.GIT_VERSION_INFO.commit)
        commit_string = ""
    ...
```

KristofferC: Are you sure? I thought it is https://github.com/JuliaCI/julia-buildkite/blob/0e1d450748e7c67cf8a95df28e9ab40761c8c6de/utilities/build_julia.sh#L36.

LilithHafner: Right, with that diff the warning would only print for untagged releases.

KristofferC: The point here is that the release is official not that it is tagged.

Drvi: Thanks both, I've updated the PR with https://github.com/JuliaLang/PackageCompiler.jl/pull/883#discussion_r1376401112

src/PackageCompiler.jl
```diff
@@ -1,6 +1,7 @@
 module PackageCompiler
 
 using Base: active_project
+using REPL
```
ericphanson, line 4: fyi it looks like REPL isn't being used in the merged version of this PR, so the dep could probably be dropped

PR #887, opened 2023-11-06
Add compat bound for stdlibs

Drvi:
plus a small cleanup (remove REPL dependency and some undefined exports)


sloede (approved):
Thanks! LGTM!

PR #889, opened 2023-11-08
Set to development version v2.1.14-pre

sloede:


KristofferC:
What advantage does this `-pre` stuff have? I have never really seen it be used.

sloede:
> What advantage does this `-pre` stuff have? I have never really seen it be used.

We use it in Trixi.jl to make sure that if you have multiple clones of a repo around and `dev`d into a number of environments, you know whether you are working with an actual release or a development version.

We had issues a (long) while back where people complained about errors with a certain release that were hard for us to reproduce, and then it turned out they were actually using `main`.

But if you find it annoying/too noisy, I can just not do it in this repo 😊

PR #890, opened 2023-11-09
Update LazyArtifacts compat to include `<0.0.1`

sharanry:


KristofferC:
This should probably be done for all stdlib entries.

(N.b, this is to work around a bug in Pkg for < 1.9 with compat on stdlibs).

Issue #783, opened 2023-02-08
precompilation is very slow with incremental=false


JeffBezanson:
`create_fresh_base_sysimage` outputs a .ji file, which is then used to precompile the environment. Because it does not have native code, each precompile process takes a very long time to start so precompiling takes ~forever.

IanButterworth:
See https://github.com/JuliaLang/PackageCompiler.jl/issues/778 it also appears to max-out all available threads, even for precompilation where no multithreading should be happening.

IMO it looks like all threads are needlessly spinning, I guess that could be happening even if 1 thread is given, slowing the process down.

Seems like a bug to me but @vtjnash said (correct me if this is an unfair summary) it was unreasonable to expect julia to behave better that this in that situation.

There was a thread in the #internals slack channel

KristofferC:
Why was this not an issue in 1.8?

JeffBezanson:
I think it was an issue on 1.8, it just got a bit worse.

PR #900, opened 2023-12-13
Make the base sysimage not include any stdlibs

KristofferC:
This is not needed since the sysimage we build on top of this will include all stdlibs that we need anyway.

Also, create a proper sysimage of the base sysimage (instead of a ji file) to improve precompilation performance of the parallel precompile we run using this sysimage.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -212,21 +212,23 @@ function get_julia_cmd()
 end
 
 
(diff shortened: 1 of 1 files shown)


KristofferC:
I think with this implementation, this will fail to include e.g. LinearAlgebra when `filter_stdlibs = false`.

JeffBezanson:
Fixes #783?

KristofferC:
It should yes, let's split this PR into two so we can get that part in directly.

PR #901, opened 2023-12-13
skip testing on 1.8

KristofferC:
This package already takes a very long time testing and testing 1.8 feels excessive since it is neither the LTS nor the latest version.


KristofferC:
Isn't 1.6 testing that?

PR #915, opened 2024-01-11
Fix git-tree-sha1 in Artifacts.toml

nhz2:
This can be fixed now. I think thanks to https://github.com/JuliaLang/Pkg.jl/pull/3349


KristofferC:
Thanks!

PR #919, opened 2024-01-19
fix lurking bug in `_collect_artifacts`

sjkelly:
`HostPlatform` was not fully qualified in the kwarg, though
this bug was never hit because the kwarg default is never used.


sloede (approved):
Thanks for the catch!

PR #922, opened 2024-02-01
Bump codecov/codecov-action from 3 to 4 + pass CODECOV_TOKEN

sloede:
Supersedes #921.

TODO:
- [x] Repo admins: please verify that an appropriate codecov token is set as the repository secret named `CODECOV_TOKEN` before merging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,9 +59,11 @@ jobs:
         with:
           coverage: ${{ matrix.coverage }}
       - uses: julia-actions/julia-processcoverage@latest
-      - uses: codecov/codecov-action@v3
+      - uses: codecov/codecov-action@v4
         with:
           file: lcov.info
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
   docs:
     runs-on: ubuntu-latest
     timeout-minutes: 30


DilumAluthge:
I can set up the token. But why do we need to upgrade to v4 of the action? Can't we just stay on v3 (which supports tokenless uploads)?

sloede:
> I can set up the token. But why do we need to upgrade to v4 of the action? Can't we just stay on v3 (which supports tokenless uploads)?

In my experience, the codecov-action has been somewhat brittle in the past (but YMMV). Therefore, in my other repos we try to stick to the latest support version. Also, all the docs are by now referring to v4-specific details, making it slightly more confusing when sticking with v3. Tokenless uploads are still supported for PRs from forks, so that should still work.

Having said that, I have no strong feelings either way here 🤷‍♂️

DilumAluthge:
Alright, I grabbed the token from the Codecov settings[^1] and I added it to this repo as a GitHub Actions Secret[^2] named `CODECOV_TOKEN`.

I'll defer to @KristofferC and @IanButterworth on the question of upgrading to v4 versus staying on v3.

[^1]: https://app.codecov.io/gh/JuliaLang/PackageCompiler.jl/settings
[^2]: https://github.com/JuliaLang/PackageCompiler.jl/settings/secrets/actions

sjkelly:
A token-based auth seems better than whatever was done before. The coverage results appear to upload correctly, so merging this. Thanks!

PR #930, opened 2024-03-19
Change naming convention of object files to follow what Base Julia does

### LINKED ISSUE #738: macOS: System image file fails consistency check on binary startup
Upon creating an app, the process goes on smoothly, untill it is time to actually call the binary.
I get the error: `ERROR: System image file failed consistency check: maybe opened the wrong version?`
```
Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin21.3.0)
  CPU: 10 × Apple M1 Pro
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, apple-m1)
  Threads: 8 on 8 virtual cores
Environment:
  JULIA_NUM_THREADS = 8
  JULIA_DEPOT_PATH = /Users/cshen/.local/julia
```

I installed julia through `juliaup`. the juliaup folder is in `~/.local/juliaup` and the julia versions in `~/.local/julia/juliaup`.
the binaries in `~/.local/juliaup/bin` are symlinked to `~/.local/bin` which is in the `PATH`.
I also have julia aliased to `julia --project=@.`

Could it be that the symlinking throws some pathing off somewhere?

gbaraldi:
Fixes #738 

I'm not sure why this makes macos broken, but given that the fix is simple I'm also not going to dig into it. 

The tests don't all pass because `fooifier` isn't there for macos, @staticfloat would need to rebuild it or we just switch to another random artifact. 

It also prints a bunch of
``` 
ld: warning: no platform load command found in '/var/folders/fs/z9fkqy2n5djg4663njw_c20m0000gn/T/jl_4TmZI5eNyL-o.a[2](text#0.o)', assuming: macOS
ld: warning: no platform load command found in '/var/folders/fs/z9fkqy2n5djg4663njw_c20m0000gn/T/jl_4TmZI5eNyL-o.a[3](text#1.o)', assuming: macOS
ld: warning: no platform load command found in '/var/folders/fs/z9fkqy2n5djg4663njw_c20m0000gn/T/jl_4TmZI5eNyL-o.a[4](text#2.o)', assuming: macOS
ld: warning: no platform load command found in '/var/folders/fs/z9fkqy2n5djg4663njw_c20m0000gn/T/jl_4TmZI5eNyL-o.a[5](text#3.o)', assuming: macOS
ld: warning: no platform load command found in '/var/folders/fs/z9fkqy2n5djg4663njw_c20m0000gn/T/jl_4TmZI5eNyL-o.a[6](text#4.o)', assuming: macOS
``` 
Which was fixed by https://github.com/JuliaLang/julia/pull/51830 but that didn't get backported into 1.10.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -226,6 +226,10 @@ function create_fresh_base_sysimage(stdlibs::Vector{String}; cpu_target::String,
     base_dir = dirname(sysimg_source_path)
     tmp_corecompiler_ji = joinpath(tmp, "corecompiler.ji")
     tmp_sys_o = joinpath(tmp, "sys-o.a")
+    # This naming convention (`sys-o.a`) is necessary to make the sysimage
+    # work on macOS.
+    # Bug report: https://github.com/JuliaLang/PackageCompiler.jl/issues/738
+    # PR: https://github.com/JuliaLang/PackageCompiler.jl/pull/930
     tmp_sys_sl = joinpath(tmp, "sys." * Libdl.dlext)
 
     compiler_source_path = joinpath(base_dir, "compiler", "compiler.jl")
@@ -640,6 +644,10 @@ function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector
 
     # Create the sysimage
     object_file = tempname() * "-o.a"
+    # This naming convention (`-o.a`) is necessary to make the sysimage
+    # work on macOS.
+    # Bug report: https://github.com/JuliaLang/PackageCompiler.jl/issues/738
+    # PR: https://github.com/JuliaLang/PackageCompiler.jl/pull/930
 
     create_sysimg_object_file(object_file, packages, packages_sysimg;
                             project,
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -225,7 +225,7 @@ function create_fresh_base_sysimage(stdlibs::Vector{String}; cpu_target::String,
     sysimg_source_path = Base.find_source_file("sysimg.jl")
     base_dir = dirname(sysimg_source_path)
     tmp_corecompiler_ji = joinpath(tmp, "corecompiler.ji")
-    tmp_sys_o = joinpath(tmp, "sys.o")
+    tmp_sys_o = joinpath(tmp, "sys-o.a")
     tmp_sys_sl = joinpath(tmp, "sys." * Libdl.dlext)
 
     compiler_source_path = joinpath(base_dir, "compiler", "compiler.jl")
@@ -639,7 +639,7 @@ function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector
     end
 
     # Create the sysimage
-    object_file = tempname() * ".o"
+    object_file = tempname() * "-o.a"
 
     create_sysimg_object_file(object_file, packages, packages_sysimg;
                             project,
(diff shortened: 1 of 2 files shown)


staticfloat:
For the `fooifier` stuff, probably best to just use a JLL like `HelloWorldC_jll` or something like that.  We don't want to actually add it as a dependency though, we want to embed the `Artifacts.toml` file directly, to continue testing the artifacts handling, so do something like just download an `Artifacts.toml` file from a JLL and embed it directly into the test suite here.

sloede (commented):
I am really happy to see this issue fixed, thanks a lot, though I truly am baffled by *how* this fix is supposed to work 😬

Just two small remarks:
1) The `ld` warning seems to appear only on macOS platforms, right? Can we maybe add an `@info` statement on macOS systems immediately before or after, informing the user that if said warning appears, it is ok to ignore?
2) The rationale for using the `-o.a` extension is very subtle. It would be imho great to add a small comment that explains why this is necessary (and/or links to this PR).

DilumAluthge:
@gbaraldi What's the status here? We just need to fix the `fooifier` problem?

DilumAluthge:
Bump @gbaraldi @staticfloat 

What needs to be done to fix CI here?

src/PackageCompiler.jl
```diff
@@ -225,7 +225,7 @@ function create_fresh_base_sysimage(stdlibs::Vector{String}; cpu_target::String,
     sysimg_source_path = Base.find_source_file("sysimg.jl")
     base_dir = dirname(sysimg_source_path)
     tmp_corecompiler_ji = joinpath(tmp, "corecompiler.ji")
-    tmp_sys_o = joinpath(tmp, "sys.o")
+    tmp_sys_o = joinpath(tmp, "sys-o.a")
```
DilumAluthge, line 228:
```suggestion
    # https://github.com/JuliaLang/PackageCompiler.jl/pull/930
    tmp_sys_o = joinpath(tmp, "sys-o.a")
```

src/PackageCompiler.jl
```diff
@@ -639,7 +639,7 @@ function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector
     end
 
     # Create the sysimage
-    object_file = tempname() * ".o"
+    object_file = tempname() * "-o.a"
```
DilumAluthge, line 642:
```suggestion
    # The `-o.a` is necessary. See https://github.com/JuliaLang/PackageCompiler.jl/pull/930
    object_file = tempname() * "-o.a"
```

src/PackageCompiler.jl
DilumAluthge, line 228:
```suggestion
    tmp_sys_o = joinpath(tmp, "sys-o.a")
    # This naming convention (`sys-o.a`) is necessary to make the sysimage
    # work on macOS.
    # Bug report: https://github.com/JuliaLang/PackageCompiler.jl/issues/738
    # PR: https://github.com/JuliaLang/PackageCompiler.jl/pull/930
```

src/PackageCompiler.jl
DilumAluthge, line 642:
```suggestion
    object_file = tempname() * "-o.a"
    # This naming convention (`-o.a`) is necessary to make the sysimage
    # work on macOS.
    # Bug report: https://github.com/JuliaLang/PackageCompiler.jl/issues/738
    # PR: https://github.com/JuliaLang/PackageCompiler.jl/pull/930
```

DilumAluthge:
> The rationale for using the `-o.a` extension is very subtle. It would be imho great to add a small comment that explains why this is necessary (and/or links to this PR).

This is done now: 93874060fc8dbddb9b7628f228b3dae3ee28736e

DilumAluthge:
Note: I have merged #967 into this PR.

DilumAluthge:
CI is all green.

PR #931, opened 2024-03-20
Allow the user to set `julia-args` for precompile execution

Drvi:
So they can run their potentially expensive precompile execution scripts with `-O0`.

Not sure if hardcoding `-O0` would be preferable... I'm open to suggestions!

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -520,7 +520,7 @@ compiler (can also include extra arguments to the compiler, like `-g`).
   for example `-O1 --check-bounds=yes`.
 
 - `precompile_execution_args::Cmd`: A set of command line options that is used in the Julia process running the
-  precompile execution file, for example `-O0`.
+  precompile execution file, default: `-O0`.
 """
 function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector{Symbol}}=nothing;
                          sysimage_path::String,
@@ -541,7 +541,7 @@ function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector
                          soname=nothing,
                          compat_level::String="major",
                          extra_precompiles::String = "",
-                         precompile_execution_args::Cmd=``
+                         precompile_execution_args::Cmd=`-O0`,
                          )
     # We call this at the very beginning to make sure that the user has a compiler available. Therefore, if no compiler
     # is found, we throw an error immediately, instead of making the user wait a while before the error is thrown.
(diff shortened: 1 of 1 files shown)


Drvi:
I guess some precompilation workloads are not particularly light computationally and could take longer without optimizations... But yeah maybe `-O0` should be the default :+1:

NHDaly:
> I guess some precompilation workloads are not particularly light computationally and could take longer without optimizations... But yeah maybe `-O0` should be the default 👍

ah true, good point. 👍

src/PackageCompiler.jl
```diff
@@ -537,6 +541,7 @@ function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector
                          soname=nothing,
                          compat_level::String="major",
                          extra_precompiles::String = "",
+                         precompile_execution_args::Cmd=``
```
NHDaly, line 544: I really think this should _default to_ `-O0`. I think that in the normal case, users will all want to set this. You're right that in some corner cases they may have big datasets and want to use O2, but i think that is the exception, not the norm.

If you are worried about that being a breaking change, we could do it in two different PRs? Otherwise, i'd put it in this one.

NHDaly (approved):
LGTM!

NHDaly:
Are the tests on Nightly unrelated?

Drvi:
I checked a couple of nighly failures and there seems to be a bounds error:
`BoundsError(a=Array{Base.Partr.taskheap, 1}(dims=(0,), mem=Memory{Base.Partr.taskheap}(0, 0x7f3fe5b3e0b0)[]), i=(0,))`

sjkelly:
This is a good idea. My hope is that propagating and adding all sorts of these configurations will be easier with, https://github.com/JuliaLang/PackageCompiler.jl/pull/929

Can we split the difference and do `-O1`?

src/PackageCompiler.jl
```diff
@@ -834,7 +843,8 @@ function create_app(package_dir::String,
                     sysimage_build_args::Cmd=``,
                     include_transitive_dependencies::Bool=true,
                     include_preferences::Bool=true,
-                    script::Union{Nothing, String}=nothing)
+                    script::Union{Nothing, String}=nothing,
+                    precompile_execution_args::Cmd=`-O0`)
```
sjkelly, line 847: Can we split the difference and do -O1?

NHDaly: Seems plausible to me, but what is that based on, @sjkelly?  My gut is that -O0 would be better for almost all snoop workloads, since the developer is already pretty strongly incentivized to keep data sizes small, and thus to have compilation-dominated workloads.

What leads you to prefer -O1?

PR #936, opened 2024-04-04
Make implicit function definition (in `julia_init.c`) an error

topolarity:
Someone more familiar with Julia than with C is not likely to recognize the significance of this error, but it's almost always a sign that you've built an object with a broken ABI somewhere.

I've also almost never seen this intentionally ignored in the wild, so I don't expect many false positives.

### DIFF
--- a/src/juliaconfig.jl
+++ b/src/juliaconfig.jl
@@ -54,7 +54,8 @@ end
 
 function cflags()
     flags = IOBuffer()
-    print(flags, "-O2 -std=gnu99")
+    print(flags, " -Werror-implicit-function-declaration")
+    print(flags, " -O2 -std=gnu99")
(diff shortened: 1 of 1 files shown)


topolarity:
fwiw, @gbaraldi would like to compile with `-Wall -Werror`

I do like that hygiene, but it might need a flag since it's not uncommon to get warnings due to headers somewhat outside your control.

sjkelly:
> fwiw, @gbaraldi would like to compile with -Wall -Werror

I think `-Wall` would be helpful.

sjkelly (approved):

topolarity:
Yeah, I agree - but let's tackle that in a separate PR to collect more input 👍

PR #938, opened 2024-04-08
Check the Julia release banner with and without the slash

DilumAluthge:
We could instead do a regex, but this seemed simpler.


LilithHafner (approved):
Duplicate of #937, but this PR comments, so it wins.

sjkelly:
Thanks!

PR #948, opened 2024-06-26
remove unofficial build warning

sjkelly:
This fixes the precompilation with Julia 1.12.
`Base.banner()` was moved to `REPL.banner()`.
This warning seems unnessecary long-term,
and more distros are making official binaries available.


KristofferC:
There is also code in place to handle non-official builds better.

KristofferC (approved):

Issue #955, opened 2024-07-30
Fix CI failures on macOS


ufechner7:
For example this one: https://github.com/JuliaLang/PackageCompiler.jl/actions/runs/10157131523/job/28086727730
and this one: https://github.com/JuliaLang/PackageCompiler.jl/actions/runs/10157131523/job/28086727940

```
$(get_julia_cmd()) --sysimage=$sysimage -e 'using Pkg; Pkg.precompile()'
```

probably has a non-zero return value when run, but without access to a macOS Julia installation that is indeed hard to impossible to debug.

ufechner7:
The first error message:
```
Testing Running tests...
  Installing known registries into `/var/folders/c9/p2llxmtj2cn82kp_6f75_gqr0000gn/T/jl_nHjCP8`
    Updating registry at `/var/folders/c9/p2llxmtj2cn82kp_6f75_gqr0000gn/T/jl_nHjCP8/registries/General.toml`
   Resolving package versions...
   Installed Example ─ v0.5.3
    Updating `/private/var/folders/c9/p2llxmtj2cn82kp_6f75_gqr0000gn/T/jl_9ryC3x/Project.toml`
  [7876af07] + Example v0.5.3
    Updating `/private/var/folders/c9/p2llxmtj2cn82kp_6f75_gqr0000gn/T/jl_9ryC3x/Manifest.toml`
  [7876af07] + Example v0.5.3
Precompiling project...
  1 dependency successfully precompiled in 1 seconds
┌ Debug: instantiating project at "/var/folders/c9/p2llxmtj2cn82kp_6f75_gqr0000gn/T/jl_9ryC3x"
└ @ PackageCompiler ~/.julia/packages/PackageCompiler/nT5sD/src/PackageCompiler.jl:570
┌ Debug: ensurecompiled: running `/Users/pkrysl/julia-1.10.4/bin/julia --color=yes --startup-file=no --pkgimages=no --sysimage=/Users/pkrysl/julia-1.10.4/lib/julia/sys.dylib -e 'using Pkg; Pkg.precompile()'`
│   JULIA_LOAD_PATH = "/var/folders/c9/p2llxmtj2cn82kp_6f75_gqr0000gn/T/jl_9ryC3x:@stdlib"
└ @ PackageCompiler ~/.julia/packages/PackageCompiler/nT5sD/src/PackageCompiler.jl:290
Precompiling project...
  1 dependency successfully precompiled in 1 seconds
[...]

PR #956, opened 2024-08-02
Do not run old Julia versions on ARM, suppress failures with Julia nightly

ufechner7:
Get rid of false positives from the continues integration tests for the ARM builds

- [x] do not run the tests for Julia 1.6, 1.7 and 1.8 on ARM, because this is not supported by the build infrastructure
- [x] run the tests on on Julia nightly, but ignore any errors, because they are expected to happen

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,8 @@ jobs:
         include:
           - os: macos-13 # Intel
             julia-version: '1.6'
+        allow_failures:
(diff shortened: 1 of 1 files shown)


ufechner7:
That the macOS tests fail is to be expected. This is https://github.com/JuliaLang/PackageCompiler.jl/issues/955 
But is it expected that the tests on 32 bit Windows fail?

DilumAluthge:
Closing in favor of #970

PR #959, opened 2024-08-17
Option to create Windows GUI app

IanET:


DilumAluthge:
Instead of specifically adding `windows_subsystem` as a kwarg, I think we should just have a more general option that allows the user to specify a custom list of additional cflags and/or ldflags.

DilumAluthge:
E.g. something like https://github.com/JuliaLang/PackageCompiler.jl/pull/920

Issue #958, opened 2024-08-02
Fix CI issues


ufechner7:
- [ ] Merge https://github.com/JuliaLang/PackageCompiler.jl/pull/956
- [ ] Fix https://github.com/JuliaLang/PackageCompiler.jl/issues/957 or decide not to support Julia 1.9 on 32 bit Windows any longer
- [ ] Fix https://github.com/JuliaLang/PackageCompiler.jl/issues/955

DilumAluthge:
Okay, so currently CI is all green on `PackageCompiler#master`. There are quite a few CI tasks remaining, including re-enabling CI on several platforms, and un-skipping several currently-broken tests.

DilumAluthge:
This is pretty much all done now. There are a few remaining tasks as listed above:
- [ ] Intermittent (non-deterministic) bug seen only in CI: #978
- [ ] TODO: Resume testing 32-bit Julia on Windows on Julia 1.11

And then there is the broader question of whether we should move CI from GitHub Actions to Buildkite: #846

Pretty much everything else is done, so I think this meta-issue can be closed now.

PR #966, opened 2024-10-15
More CI tweaks, and temporarily disable some more CI jobs

DilumAluthge:
Ref #958


DilumAluthge:
Okay, CI is all green now (although note that we have temporarily disabled quite a few jobs).

PR #967, opened 2024-10-15
CI: Resume testing on Intel macOS

DilumAluthge:
Targets `gb/macos-is-odd` (#930).

---

See also:
- #738
- #930
- #958


DilumAluthge:
I have merged this into #930.

PR #972, opened 2024-10-16
macOS: If using Xcode CLT >= 15, add `-Wl,-ld_classic` in `create_sysimg_from_object_file()`

DilumAluthge:
Replaces #935 (closes #935)
Reference #738

---

This is based on #935. All credit goes to @PhilReinhold for providing the fix and @jayscook for opening the original PR (#935).

This PR differs slightly from #935. #935 used a try-catch. This PR uses `ignorestatus()` to ignore non-zero exit codes from running the `$cc --version` command, and uses `isnothing()` to handle the case where the regex does not match. Also, this PR factors out the "detect CLT major version" functionality out into a separate helper function.

Also, this PR now addresses https://github.com/JuliaLang/PackageCompiler.jl/pull/935#discussion_r1554017266

We do not add these flags for CLT versions <= 14.


DilumAluthge:
This PR doesn't yet address https://github.com/JuliaLang/PackageCompiler.jl/pull/935#discussion_r1554017266, but I'll look into it further.

DilumAluthge:
Okay, https://github.com/JuliaLang/PackageCompiler.jl/pull/935#discussion_r1554017266 is now addressed by the combination of c9a4b52e4352197a97489950d5daf08b45c02f4a and adf57377f4952aa09b1c00dce642f70090f0361e.

DilumAluthge:
Specifically, we now run `$cc --version`, and we use that output to determine whether or not the compiler is the Clang from Xcode.

DilumAluthge:
It's not entirely clear to me if we still need this PR, because #930 seems to have fixed #738, at least in the macOS CI on this repo.

src/xcode.jl
```diff
@@ -0,0 +1,54 @@
+# Get the compiler command from `get_compiler_cmd()`, and run `cc --version`, and parse the
+# output to determine whether or not the compiler is an Xcode Clang.
+#
+# The return value is a NamedTuple of the form (; b, ver)
+# b = a Bool that is true iff he compiler is an Xcode Clang.
```
DilumAluthge, line 5:
```suggestion
# b = a Bool that is true iff the compiler is an Xcode Clang.
```

src/xcode.jl
```diff
@@ -0,0 +1,54 @@
...
+    if is_xcode_app || is_xcode_clt
+        m = match(r"\nApple clang version ([0-9\.]*?) ", output)
+        if isnothing(m)
+            @warn "Could not determine the version of the Xcode Command Line Tools"
+            (; b=false, ver=nothing)
+        end
+        ver_str = strip(m[1])
+        ver = tryparse(VersionNumber, ver_str)
+        if isnothing(ver)
+            @warn "Could not determine the version of the Xcode Command Line Tools"
```
DilumAluthge, line 28:
```suggestion
            @warn "Could not determine the version of the Xcode Command Line Tools" ver_str
```

src/xcode.jl
```diff
@@ -0,0 +1,54 @@
...
+end
+
+# Return true iff the compiler is Xcode Clang AND the Xcode CLT version is >= 15.
+#
+# If the user sets the JULIA_PACKAGECOMPILER_XCODE_CLT_MAJOR_VERSION environment variable,
+# we skip our attempt at auto-detection, and instead use whatever value the user gave us.
+function _is_xcode_clt_and_is_gte_xcode_15()
+    str = strip(get(ENV, "JULIA_PACKAGECOMPILER_XCODE_CLT_MAJOR_VERSION", ""))
+    ver_int = tryparse(Int, str)
+    (ver_int isa Int) && return ver_int
```
DilumAluthge, line 46:
```suggestion
    (ver_int isa Int) && return (ver_int >= 15)
```

jayscook:
Thanks for putting this together @DilumAluthge!

DilumAluthge:
@jayscook Thank you for putting the original PR together (and thank you to @PhilReinhold for writing the original code).

DilumAluthge:
@topolarity Could you take a look at this PR, and let me know if it addresses your concerns in https://github.com/JuliaLang/PackageCompiler.jl/pull/935#discussion_r1554017266 and https://github.com/JuliaLang/PackageCompiler.jl/pull/935#discussion_r1552442374?

topolarity:
Sure thing! Looks like the right direction, but I won't be able to take a close look until Monday

DilumAluthge:
Sounds good, thanks man!

DilumAluthge:
@sjkelly I see you also reviewed the original PR (#935), so I've requested your review on this one.

PR #970, opened 2024-10-16
Run macOS CI on both Intel and Apple Silicon

DilumAluthge:
This PR does the following:
1. Run macOS CI on both Intel and Apple Silicon
2. Fix a CI config bug regarding 32-bit vs 64-bit Julia

#973 was merged into this PR.

---

For reference, in GitHub Actions:
- `runs-on: macos-13` = Intel.
- `runs-on: macos-14` = Apple Silicon.

PR #973, opened 2024-10-16
CI: Fix a CI config bug regarding 32-bit vs 64-bit Julia

DilumAluthge:


DilumAluthge:
I have merged this into #970.

PR #982, opened 2024-10-20
CI: Run a single job on Julia nightly (but skip the `incremental=false` tests on Julia 1.12)

DilumAluthge:
1. We intentionally put nightly CI into a separate YAML file (`ci.nightly.yml` versus the main `ci.yml` file). This allows us to enable and disable the nightly CI workflow from the GitHub web UI, without having any impact on the main CI.
2. We intentionally only run one CI job on Julia nightly, in order to conserve CI resources.
3. This CI only runs on Julia nightly. It doesn't run on Julia pre-releases. For CI on Julia pre-releases, see the main CI workflow file.

### DIFF
--- a/.github/workflows/ci.nightly.yml
+++ b/.github/workflows/ci.nightly.yml
@@ -32,8 +32,8 @@ jobs:
           # 2. We intentionally only run one CI job on Julia nightly,
           #    in order to conserve CI resources.
           # 3. This CI only runs on Julia nightly. It doesn't run on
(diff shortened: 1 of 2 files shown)


DilumAluthge:
CI is failing on Julia nightly, which is obviously not good, but at least we have a CI job now. If the red :X gets too distracting, we can easily disable/enable the "Nightly CI" workflow from the GitHub web UI (this page: https://github.com/JuliaLang/PackageCompiler.jl/actions/workflows/ci.nightly.yml) without affecting main CI.

PR #983, opened 2024-10-20
CI: Test both `incremental={true,false}` on "slow CI"; also, stop considering Apple Silicon macOS as "slow CI"

DilumAluthge:
In my recent CI experiments, I've discovered some bugs that only occur in one branch of `incremental={true,false}`. So I think we should try to exercise both of those branches in CI, even if we are in "slow CI".

### DIFF
--- a/.github/workflows/ci.nightly.yml
+++ b/.github/workflows/ci.nightly.yml
@@ -32,8 +32,8 @@ jobs:
           # 2. We intentionally only run one CI job on Julia nightly,
           #    in order to conserve CI resources.
           # 3. This CI only runs on Julia nightly. It doesn't run on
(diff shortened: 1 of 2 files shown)


DilumAluthge:
I did a bit of Git archeology. As far as I can tell, the `is_slow_ci` flag was introduced before we had Apple Silicon. I wonder if we mahbe had Linux aarch64 CI back then, and that's what the `is_slow_ci` flag was meant to cover.

AFAICT, the Apple Silicon GitHub Actions CI GitHub-hosted runners on this repo seem to run the CI jobs pretty quickly.

So I don't think Apple Silicon should count as "slow CI".

Issue #976, opened 2024-10-17
Regression in Julia 1.11: Cannot create fresh base sysimage (`incremental=false`) containing Pkg


DilumAluthge:
## MWE

```julia
import PackageCompiler

sysimage_stdlibs = ["Pkg"]

base_sysimage = PackageCompiler.create_fresh_base_sysimage(
    sysimage_stdlibs;
    cpu_target = PackageCompiler.default_app_cpu_target(),
    sysimage_build_args = ``,
)
```

Expected behavior: no exception is thrown.

Actual behavior:

```
✔ [01m:35s] PackageCompiler: creating compiler .ji image (incremental=false)
⣄ [00m:55s] PackageCompiler: compiling fresh sysimage (incremental=false)error during bootstrap:
⠇ [00m:55s] PackageCompiler: compiling fresh sysimage (incremental=false)

LoadError("/var/folders/lp/2fm0z_wn3wqf__vj6lc1tcgh0000gq/T/jl_QNNqc5/sysimage_packagecompiler_fdb8261a-8ca⠋ [00m:55s] PackageCompiler: compiling fresh sysimage (incremental=false)57, LoadError("/myjuliapath/julia/1.11.0/share/julia/stdlib/v1.11/Pkg/src/Pkg.jl", 3, LoadError("/myjuliapath/julia/1.11.0/share/julia/stdlib/v1.11/Pkg/src/precompile.jl", 101, Base.IOError("could not spawn setenv(`'' a registries/Registry.tar.gz -tgzip registries/Registry.tar`,[... long list of environment variables ...]): permission denied (EACCES)", -13))))

_spawn_primitive at ./process.jl:140
[... very long stack trace ...]

✖ [00m:56s] PackageCompiler: compiling fresh sysimage (incremental=false)
[...]

PR #987, opened 2024-10-29
do not put project stdlibs into the base (non-incremental) sysimage

KristofferC:
Fixes https://github.com/JuliaLang/PackageCompiler.jl/issues/976

There is no need to do this, the stdlibs used by the project will be included when the final sysimage is built

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -594,7 +594,7 @@ function create_sysimage(packages::Union{Nothing, Symbol, Vector{String}, Vector
         if base_sysimage !== nothing
             error("cannot specify `base_sysimage`  when `incremental=false`")
         end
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Just FYI, a lot of the tests for non-incremental sysimages are failing on CI (due to https://github.com/JuliaLang/PackageCompiler.jl/issues/976), so I've disabled them and marked them as broken.

So, if you want to see the impact of this PR, you'll likely need to re-enable those tests in your PR branch.

KristofferC:
Removed all the test filtering for 1.11 and 1.12.

DilumAluthge:
Nightly (Julia 1.12) CI is failing with what seems to be a new bug (i.e. not the same bug as #976). I've opened a new issue: #989

KristofferC:
Agree

PR #988, opened 2024-10-29
Test suite: enable another non-incremental test on Julia 1.12

DilumAluthge:
Targets `kc/stdlibs` (#987).

It is best to enable the "Hide whitespace" checkbox when you view this PR's diff.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -231,26 +231,15 @@ end
 
     # Test creating an empty sysimage
     if !is_slow_ci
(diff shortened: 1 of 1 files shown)


DilumAluthge:
@KristofferC We probably also want this as part of #987/

DilumAluthge:
Resolved by https://github.com/JuliaLang/PackageCompiler.jl/pull/987/commits/ebd57d1e0b59edd4ca142a6b64c447d3553ad5a9 (which is part of #987).

Issue #989, opened 2024-10-29
Regression in Julia 1.12: Creating fresh base sysimage (`incremental=false`) fails with `error during bootstrap`


DilumAluthge:
## The error

Here is what the error message itself looks like:

> ✔ [01m:33s] PackageCompiler: creating compiler .ji image (incremental=false)
> ⡆ [00m:06s] PackageCompiler: compiling fresh sysimage (incremental=false)error during bootstrap:
> LoadError(at "/var/folders/lp/2fm0z_wn3wqf__vj6lc1tcgh0000gq/T/jl_YBjqXM/sysimage_packagecompiler_a12bf01a-97ce-11ef-0ebb-19685e72afd2.jl" line 3: LoadError(at "Base.jl" line 287: ErrorException("File \"/var/folders/lp/2fm0z_wn3wqf__vj6lc1tcgh0000gq/T/jl_YBjqXM/sysimage_packagecompiler_a12bf01a-97ce-11ef-0ebb-19685e72afd2.jlbuild_h.jl\" not found")))

<details>

<summary>
Here is the full stacktrace (click to expand):
</summary>

```
Precompiling all packages...
  4 dependencies successfully precompiled in 3 seconds. 26 already precompiled.
+ ./julia --startup-file=no --project=../PackageCompiler.jl -e 'import Pkg; Pkg.precompile()'
+ ./julia --startup-file=no --project=../PackageCompiler.jl ../mwe.jl
✔ [01m:33s] PackageCompiler: creating compiler .ji image (incremental=false)
⡆ [00m:06s] PackageCompiler: compiling fresh sysimage (incremental=false)error during bootstrap:
LoadError(at "/var/folders/lp/2fm0z_wn3wqf__vj6lc1tcgh0000gq/T/jl_YBjqXM/sysimage_packagecompiler_a12bf01a-97ce-11ef-0ebb-19685e72afd2.jl" line 3: LoadError(at "Base.jl" line 287: ErrorException("File \"/var/folders/lp/2fm0z_wn3wqf__vj6lc1tcgh0000gq/T/jl_YBjqXM/sysimage_packagecompiler_a12bf01a-97ce-11ef-0ebb-19685e72afd2.jlbuild_h.jl\" not found")))
⠇ [00m:06s] PackageCompiler: compiling fresh sysimage (incremental=false)ijl_errorf at /mydir/bisect-pc/julia/src/rtutils.c:77
[...]

PR #993, opened 2024-10-29
CI: Add the `pre` version, but keep it commented out for now (since there's not currently a prerelease available)

DilumAluthge:


DilumAluthge:
The nightly CI failure is https://github.com/JuliaLang/PackageCompiler.jl/issues/989

PR #995, opened 2024-10-29
Set `JULIA_PROJECT` when running the subprocess in `ensurecompiled()`

DilumAluthge:
Suppose that you have two different projects, where one project contains the packages that will go into the sysimage, and the other project is a "build project".

As an example, consider the [`examples/MyLib`](https://github.com/JuliaLang/PackageCompiler.jl/tree/master/examples/MyLib) folder in this repo:
- The `examples/MyLib/Project.toml` is the "main project" that contains the packages that will go into the sysimage.
- The `examples/MyLib/build/Project.toml` is the "build project". The packages in the "build project" will not go into the sysimage. In this example, the "build project" contains `Libdl` and `PackageCompiler`, but neither of those will go into the sysimage.

When the `ensurecompiled()` function is run, we want to make sure that it precompiles the "main project", not the "build project". That's what this PR does.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -299,8 +299,9 @@ function ensurecompiled(project, packages, sysimage)
     # TODO: Only precompile `packages` (should be available in Pkg 1.8)
     cmd = `$(get_julia_cmd()) --sysimage=$sysimage -e 'using Pkg; Pkg.precompile()'`
     splitter = Sys.iswindows() ? ';' : ':'
-    @debug "ensurecompiled: running $cmd" JULIA_LOAD_PATH = "$project$(splitter)@stdlib"
-    cmd = addenv(cmd, "JULIA_LOAD_PATH" => "$project$(splitter)@stdlib")
+    JULIA_LOAD_PATH = "$project$(splitter)@stdlib"
+    @debug "ensurecompiled: running $cmd" JULIA_LOAD_PATH
+    cmd = addenv(cmd, "JULIA_LOAD_PATH" => JULIA_LOAD_PATH, "JULIA_PROJECT" => project)
     run(cmd)
     return
 end


KristofferC:
I would have think this is already the case from the existing code? Do you have something like an MWE where it isn't?

```
julia> Base.active_project()
"/home/kc/JuliaPkgs/PackageCompiler.jl/examples/MyLib/build/Project.toml"

julia> project = ".."
".."

julia> run(addenv(`$(Base.julia_cmd()) -e '@show Base.load_path()'`, "JULIA_LOAD_PATH" => "$project:@stdlib"));
Base.load_path() = ["/home/kc/JuliaPkgs/PackageCompiler.jl/examples/MyLib/Project.toml", "/home/kc/.julia/juliaup/julia-1.10.6+0.x64.linux.gnu/share/julia/stdlib/v1.10"]
```

DilumAluthge:
Good question. I haven't been able to come up with a MWE. I'm encountering some kind of problem (that I thought was `JULIA_PROJECT`-related) when I try to move the `build-mylib` test into the PackageCompiler test suite (instead of having it as a separate CI job). That's what I've been trying in #991.

DilumAluthge:
Okay, so I do have an example where things don't seem to work unless I have this `JULIA_PROJECT` change.

In #991, I'm working on moving the `build-mylib` test into the regular test suite (and thus get rid of the separate `build-mylib` CI job).

In that PR, if I don't have this `JULIA_PROJECT` change, then CI fails (see https://github.com/JuliaLang/PackageCompiler.jl/pull/991/commits/8308cf808ed06a39d630919d4ae84c13922226a4, e.g. [this example log](https://github.com/JuliaLang/PackageCompiler.jl/actions/runs/12003433149/job/33456879494)) with the following:

```
ERROR: The following 1 direct dependency failed to precompile:

PackageCompiler [9b87118b-4619-50d2-8e1e-99f35a4d4d9d]

Error: Missing source file for PackageCompiler [9b87118b-4619-50d2-8e1e-99f35a4d4d9d
```

However, when I add this `JULIA_PROJECT` change, then CI passes (see https://github.com/JuliaLang/PackageCompiler.jl/pull/991/commits/f4bdaaf0bb2a79f60406361753d4060013c468c5).

PR #996, opened 2024-10-31
`create_fresh_base_sysimage()`: Pass `""` as the BUILDROOT for Julia 1.12+

DilumAluthge:
Fixes #989.

This fixes the immediate bug in #989, which is that Base.jl incorrectly assumes that our first positional argument is the value of the `BUILDROOT`. The fix is to pass a second positional argument `""`, which is then taken by Base.jl to be the `BUILDROOT`.

This bugfix PR will eventually be followed up by #997, which is a larger PR that will actually take advantage of the new BuildSettings mechanism added to Julia 1.12 in https://github.com/JuliaLang/julia/pull/54387

PR #997, opened 2024-11-01
Use BuildSettings for fresh base sysimages on Julia 1.12+

DilumAluthge:
Ref https://github.com/JuliaLang/PackageCompiler.jl/issues/989#issuecomment-2450991016

Follow-up to #996.

This uses the BuildSettings mechanism added in JuliaLang/julia#54387. PackageCompiler now always creates an empty fresh base sysimage and adds any requested stdlibs later. Because Julia parses `INCLUDE_STDLIBS = ""` as a package with an empty name, `Base` is used as an already-loaded no-op sentinel.

On Julia 1.11 and earlier, the existing `sysimg.jl` rewrite remains in place.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -332,14 +332,6 @@ function warn_if_sysimage_too_big(sysimage_path::String)
 end
 
 
-function rewrite_sysimg_jl_only_needed_stdlibs()
-    sysimg_source_path = Base.find_source_file("sysimg.jl")
-    sysimg_content = read(sysimg_source_path, String)
-    # replaces the hardcoded list of stdlibs in sysimg.jl with an empty list
-    # TODO: Use the mechanism in https://github.com/JuliaLang/PackageCompiler.jl/pull/997
-    return replace(sysimg_content, r"stdlibs = \[(.*?)\]"s => "stdlibs = []")
-end
-
 function create_fresh_base_sysimage(; cpu_target::String, sysimage_build_args::Cmd)
     tmp = mktempdir()
     sysimg_source_path = Base.find_source_file("sysimg.jl")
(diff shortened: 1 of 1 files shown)


src/PackageCompiler.jl
```diff
@@ -254,16 +262,43 @@ function create_fresh_base_sysimage(stdlibs::Vector{String}; cpu_target::String,
...
+
+                # The second positional argument `""` is to pass BUILDROOT="" to Base.jl.
+                # If we don't have the "", then Base.jl will assume that one of our other
+                # arguments is the BUILDROOT, which obviously is incorrect.
+                # https://github.com/JuliaLang/PackageCompiler.jl/issues/989
+                #
+                # The --build-settings argument is to pass our BuildSettings file.
+                # BuildSettings is only available in Julia 1.12+
+                # https://github.com/JuliaLang/julia/pull/54387
+                buildsettings_args = `$new_sysimage_source_path "" --build-settings $(build_settings_source_path)`
```
KristofferC, line 287: I would move all the logic for rewriting the sysimage.jl file into the else branch to show that it is dead on newer julia versions and just use the `Base.find_source_file("sysimg.jl")` as the argument here to show nothing weird is going on.

JeffBezanson:
```suggestion
                buildsettings_args = `$new_sysimage_source_path "" --buildsettings $(build_settings_source_path)`
```

src/PackageCompiler.jl
```diff
@@ -392,7 +384,17 @@ function create_fresh_base_sysimage(; cpu_target::String, sysimage_build_args::C
         spinner = TerminalSpinners.Spinner(msg = "PackageCompiler: compiling fresh sysimage (incremental=false)")
         TerminalSpinners.@spin spinner begin
             # Use the compiler sysimage to create sys.ji
-            new_sysimage_content = rewrite_sysimg_jl_only_needed_stdlibs()
+            sysimg_content = read(sysimg_source_path, String)
+            @static if VERSION >= v"1.12.0-DEV.1617"
+                # `INCLUDE_STDLIBS = ""` is parsed as a package with an empty name.
+                # Base is already loaded, so requiring it is a no-op.
+                new_sysimage_content =
+                    "Core.eval(Base.BuildSettings, :(INCLUDE_STDLIBS = \"Base\"))\n" * sysimg_content
```
KristofferC, line 392: Funny hack

KristofferC:
I don't think this is needed anymore since we never build a fresh sysimage with stdlibs anymore.

PR #1005, opened 2024-11-21
Bump `julia` [compat] to 1.10 (and drop support for 1.6 through 1.9)

KristofferC:
The CI matrix has ballooned up to a ridiculous number of combinations at this point. First step to reduce it seems to be to stop supporting and testing unsupported julia releases.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -30,24 +30,12 @@ const is_gha_ci = tryparse(Bool, get(ENV, "GITHUB_ACTIONS", "")) === true
 #    macOS runners seem to be quite fast.)
 const is_slow_ci = is_ci && Sys.ARCH == :aarch64 && !Sys.isapple()
 
-const is_julia_1_6 = VERSION.major == 1 && VERSION.minor == 6
-const is_julia_1_9 = VERSION.major == 1 && VERSION.minor == 9
-
 if is_ci || is_gha_ci
     @info "This is a CI job" Sys.ARCH VERSION is_ci is_gha_ci
 end
 
 if is_slow_ci
     @warn "This is \"slow CI\" (defined as any non-macOS CI running on aarch64). Some tests will be skipped or modified." Sys.ARCH
-end
-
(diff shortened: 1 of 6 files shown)


DilumAluthge:
Dropping some old CI jobs seems okay. Is it necessary to bump the Julia compat? We could just leave the Julia compat as is, and if new PackageCompiler releases continue to work on old Julia releases, that seems fine, right?

I think it'd be nice to at least keep compatibility with 1.6 for now, since it was the LTS for a long time, and there are likely still people using it.

test/runtests.jl
```diff
@@ -117,7 +105,7 @@ end
             @info "starting: create_app testset" incremental filter
             tmp_app_source_dir = joinpath(tmp, "MyApp")
             cp(app_source_dir, tmp_app_source_dir)
-            if is_gha_ci && (is_julia_1_6 || is_julia_1_9)
+            if is_gha_ci
```
DilumAluthge, line 108: You'd want to remove this entire `if`-`end` block.

KristofferC:
People can still use PackageCompiler with 1.6.

I think getting rid of a bunch of old version checks to simplify the code is worth it. For example I think https://github.com/JuliaLang/PackageCompiler.jl/blob/8d6fce040a41432e2504b9be9eca797ee9ab1f58/src/PackageCompiler.jl#L1169 could be removed now as well.

DilumAluthge (approved):
Personally, I don't think that the relatively minor (IMO) code simplifications are worth dropping support for 1.6. My thinking is that 1.6 was the LTS for a long time, so people are probably going to keep using 1.6 for a while, and it would be nice for those users to keep getting PackageCompiler improvements.

That all being said, I don't feel particularly strongly about this, so I won't stand in the way.

We should make the next PackageCompiler release 2.2.0. That way, if we do end up needing to backport a bug fix, we can do so in the 2.1.x release series.

DilumAluthge (approved):

PR #1006, opened 2024-11-21
CI: Always run tests with code coverage enabled (remove the coverage-disabled CI jobs)

KristofferC:
I don't fully understand what this comment means (cc @IanButterworth) but running all tests with and without coverage seems a bit bananas so I'll check what happens if only coverage is used.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,6 @@ jobs:
           - macos-14 # macos-14 = Apple Silicon.
         coverage:
           - 'true'
-          - 'false' # needed for Julia 1.9+ to test from a session using pkgimages
         exclude:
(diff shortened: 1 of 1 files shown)


IanButterworth:
IIRC this pre-dated https://github.com/JuliaLang/julia/pull/52123 so can be removed now.

DilumAluthge:
> IIRC this pre-dated [JuliaLang/julia#52123](https://github.com/JuliaLang/julia/pull/52123) so can be removed now.

And indeed, the `codecov/project` commit status shows that the coverage on this PR is the same as the coverage on master.

So we'll be good to go here once the last CI job passes.

DilumAluthge:
CI is all green here.

Codecov statuses are also both green here.

Issue #918, opened 2024-01-19
Need to update Windows compilers for Julia 1.11


KristofferC:
The current one we use spits out a bunch of

```
2024-01-15T16:40:05.8024498Z Warning: .drectve `-exclude-symbols:jfptr_only_42205 ' unrecognized
2024-01-15T16:40:05.8052985Z Warning: .drectve `-exclude-symbols:julia_eval_41586 ' unrecognized
2024-01-15T16:40:05.8083804Z Warning: .drectve `-exclude-symbols:jfptr_eval_41587 ' unrecognized
2024-01-15T16:40:05.8114625Z Warning: .drectve `-exclude-symbols:jfptr_domath_65456 ' unrecognized
2024-01-15T16:40:05.8144650Z Warning: .drectve `-exclude-symbols:julia_peek_45894 ' unrecognized
```

on master CI when creating the sysimage. @gbaraldi suggested to try use https://github.com/mstorsjo/llvm-mingw/.

DilumAluthge:
Just out of curiosity, are these warnings fatal and/or do they result in an un-usable sysimage?

(I think we should still do this task either way; I'm just trying to figure out how high of a priority it is.)

smitha999:
Am getting similar errors while trying to do an FMU Export from Julia

smitha999:
Warning: .drectve `-exclude-symbols:julia_copy_26305 ' unrecognized
│  Warning: .drectve `-exclude-symbols:jfptr_copy_26306 ' unrecognized
│  Warning: .drectve `-exclude-symbols:jfptr_ntuple_25526 ' unrecognized
│  Warning: .drectve `-exclude-symbols:jfptr__memory_offset_28906 ' unrecognized
│  Warning: .drectve `-exclude-symbols:jfptr_throw_boundserror_23456 ' unrecognized
│  Warning: .drectve `-exclude-symbols:julia_throw_boundserror_23449 ' unrecognized
│  Warning: .drectve `-exclude-symbols:jfptr_throw_boundserror_23441 ' unrecognized
│  .
    .
│  Warning: .drectve `-exclude-symbols:jfptr_throw_boundserror_23279 ' unrecognized
│  Warning: .drectve `-exclude-symbols:"jfptr_setindexNOT._24863" ' unrecognized
│  Warning: .drectve `-exclude-symbols:julia_has_safe_def_35574 ' unrecognized

inkydragon:
[...]

PR #1012, opened 2024-11-27
update Windows compilers

KristofferC:
It would be better to link to an official URL but those are in 7z format and Artifacts require .tar.gz so I had to repack it (similar to what was earlier done). 

Fixes https://github.com/JuliaLang/PackageCompiler.jl/issues/918


KristofferC:
Actually, I think the problem is something else (no file extension -> .exe by the compiler).

KristofferC:
The previous compilers where also repackaged (also by me) so I think this is fine.

PR #1041, opened 2025-04-15
make sure 1 thread is used in julia processes

KristofferC:
ref, https://discourse.julialang.org/t/packagecompiler-hangs-while-compiling-nonincremental-system-image-on-julia-v1-11/128099/14

Issue #990, opened 2024-10-29
create_app never finishes on Julia v1.11.1


juanromerohb:
The same app I was successfully compiling in Julia v.1.10.6 under 15 minutes no longer compiles after updating Julia to v1.11.1.

It stucks +20 minutes on

`⠸ [03m:30s] PackageCompiler: compiling nonincremental system image`

The difference I noted was that before `PackageCompiler: compiling fresh sysimage (incremental=false)`, it throws me thousands of warnings like these

```
Warning: .drectve `-exclude-symbols:"julia_IRInterpretationState_54964.reloc_slot" ' unrecognized
Warning: .drectve `-exclude-symbols:"julia_YY.tarjanNOT.YY.379_64959.reloc_slot" ' unrecognized
Warning: .drectve `-exclude-symbols:jl_fvar_count_5 ' unrecognized
Warning: .drectve `-exclude-symbols:jl_fvar_ptrs_5 ' unrecognized
Warning: .drectve `-exclude-symbols:jl_clone_slots_5 ' unrecognized
Warning: .drectve `-exclude-symbols:jl_clone_idxs_5 ' unrecognized
Warning: corrupt .drectve at end of def file
Warning: .drectve `-exclude-symbols:jl_shard_tables ' unrecognized
Warning: .drectve `-exclude-symbols:jl_pgcstack_func_slot ' unrecognized
Warning: .drectve `-exclude-symbols:jl_pgcstack_key_slot ' unrecognized
Warning: .drectve `-exclude-symbols:jl_tls_offset ' unrecognized
Warning: .drectve `-exclude-symbols:jl_ptls_table ' unrecognized
Warning: .drectve `-exclude-symbols:jl_small_typeof ' unrecognized
```

DilumAluthge:
1. Which operating system are you using?
2. Do you have an MWE?

juanromerohb:
> 1. Which operating system are you using?
> 2. Do you have an MWE?

I'm working on Windows 11 using Visual Studio Code with the Julia REPL extension..

A MWE I just tried starts by generating an empty Julia application:

```
using Pkg
Pkg.generate("MyApp")
```
Next, I modified the `MyApp/src/MyApp.jl` file to create a basic "Hello World" program:

```
module MyApp

function julia_main()::Cint
    println("Hello, World!")
end

end
```

Then, I run

```
using PackageCompiler
create_app("MyApp", "MyAppCompiled")
```

[...]

PR #1042, opened 2025-04-22
Fix hanging system image creation if multiple threads are set

### LINKED ISSUE #963: PackageCompiler never ends on julia 1.11.0 with sysimage_build_args=Cmd(["--threads=6"])
The following command never finishes on julia 1.11.0 (worked on julia 1.10.5)
```julia
using PackageCompiler
create_sysimage(;sysimage_path="kk.so", precompile_statements_file="../kk.trace", sysimage_build_args=Cmd(["--threads=6"]))
```

The timer keeps on counting forever (CPU does not show any work done)
` [09m:10s] PackageCompiler: compiling incremental system image`

The fix is not specifying `--threads=6` on sysimage_build_args

PatrickHaecker:
Make sure, that the final system image is built single-threaded and override any values set by "-t", "--threads" in sysimage_build_args` or provided via JULIA_NUM_THREADS.
This is needed until the underlying bug is fixed (see #963 and especially #990 containing a `git bisect` to the commit introducing the problem)

References #963 and #990 (but does not fix the underlying bug)

Supersedes #1041 (closes #1041).

Thanks a lot to @KristofferC for the close hit with #1041 limiting down the search area enormously and also to everyone reporting or contributing to the discussions around #963 and #990, especially to @dpinol, @juanromerohb,  @matthewgcooper and last, but certainly not least for a lot of patience with `git bisect`, to @paulmelis.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -465,8 +465,8 @@ function create_sysimg_object_file(object_file::String,
     # Make sure, that the final system image is built single-threaded and
     # override any values set by "-t", "--threads" in `sysimage_build_args`
     # or provided via JULIA_NUM_THREADS.
-    # This is needed until the underlying bug is fixed (see #963 and especially
-    # #990 containing a `git bisect` to the commit introducing the problem)
+    # This is needed until the underlying bug is fixed (see https://github.com/JuliaLang/PackageCompiler.jl/issues/963 and especially
+    # https://github.com/JuliaLang/PackageCompiler.jl/issues/990 containing a `git bisect` to the commit introducing the problem)
     cmd = `$(get_julia_cmd()) --cpu-target=$cpu_target $sysimage_build_args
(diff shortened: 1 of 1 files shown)


PatrickHaecker:
I have a wish:
As this issue already sparked some discussions on Discourse, can we have this merged and released quickly if there are no issues?
This would also help me, because I could skip creating an internal workaround, which I am lacking a bit the time for because I rather spent the time to fix the problem upstream.

donovaly:
Thanks. I can confirm it fixes the problem, see also my [post in Discourse](https://discourse.julialang.org/t/packagecompiler-hangs-while-compiling-nonincremental-system-image-on-julia-v1-11/128099/52?u=uwestoehr)

j-hayes:
I also tested this with my multithreaded code and it worked

DilumAluthge:
I'm working on releasing 2.2.1, which will include this.

PR #1054, opened 2025-08-12
Remove sysimg creation workaround.

### LINKED ISSUE #839: Is `create_sysimage_workaround` still needed?
According to the associated comment,
https://github.com/JuliaLang/PackageCompiler.jl/blob/7f0607c990003099969f449c28ad2aeffd0671f4/src/PackageCompiler.jl#L1047-L1050
the function `create_sysimage_workaround` should not be necessary anymore once https://github.com/JuliaLang/julia/pull/37530 has been merged (which it was, approximately 2 years ago).

Since it seems like this workaround increases compile times, is it time to retire the workaround? Or is it still needed and only the comment needs updating?

maleadt:
https://github.com/JuliaLang/julia/pull/37530 is part of 1.8+, and PackageCompiler.jl requires Julia 1.10.

Closes https://github.com/JuliaLang/PackageCompiler.jl/issues/839.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1097,10 +1097,29 @@ function create_library(package_or_project::String,
     compat_file = get_library_filename(lib_name; version, compat_level)
     soname = (Sys.isunix() && !Sys.isapple()) ? compat_file : nothing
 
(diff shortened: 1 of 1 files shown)


KristofferC:
👍 CI failure is probably unrelated.

PR #1056, opened 2025-08-21
Fix `git-tree-sha1` for i686-w64-mingw32 GCC artifact

topolarity:
Not sure how this was wrong, but saw this in CI and verified locally with `ArtifactUtils.jl`


KristofferC:
Thanks!

PR #1073, opened 2025-11-19
Add a `create_distribution` entry point

KristofferC:
This creates basically a new julia "distribution" with a custom set of packages in it.

Right now:

- packages get added to the sysimage
- packages are treated as stdlibs so they can be loaded via `using Package` without a corresponding project file
- packages are also treated as stdlibs by Pkg
- package's source code is not copied, shims are added so that Pkg will properly treat it as an stdlib

TODO:

- [x] test
- [x] see how extensions are treated


src/PackageCompiler.jl
```diff
@@ -979,6 +991,92 @@ function create_app(package_dir::String,
...
+- `force::Bool=false`: Overwrite `dist_dir` if it exists.
+- `cpu_target::String=default_app_cpu_target()`: CPU target used when compiling the sysimage.
+- `include_lazy_artifacts::Bool=false`: If `true`, lazy artifacts referenced by dependencies are bundled.
+- `sysimage_build_args::Cmd=```: Additional flags for the Julia process building the sysimage.
+- `include_transitive_dependencies::Bool=true`: If `true`, include transitive dependencies in the sysimage.
+- `include_preferences::Bool=true`: Bundle package preferences into `share/julia/LocalPreferences.toml`.
+- `script::Union{Nothing,String}=nothing`: Optional script executed while generating the sysimage.
+- `copy_globs::Vector{String}=String[]`: Glob patterns for copying package files to the stdlib directory.
+  Patterns are relative to each package root and apply to all packages in the distribution.
+  Example: `["assets/**", "data/**"]` copies assets and data directories for all packages.
```
bauglir: When specifying globs, if one forgets to include `"src/**"`, you end up with a distribution that does not work because none of the packages can be loaded. I'm not 100% sure if this is true in general, but in my testing when I messed this up nothing would work.

I'm wondering if making the stub generation conditional on whether the `src` folder was copied over. That way there is no extra configuration to be done, although I guess there could be an argument to explicitly disable it, and when someone forgets to add `"src/**"` to the globs the distribution still ends up working.

bauglir: Perhaps the stub generation is even conditional on whether the package entrypoint is already in the target location or not.

KristofferC: I don't think this is true unless the package somehow reads file during runtime from `src`. Otherwise, packages should be in the sysimage and not require any bundled sources. Do you have an example of this with some simple packages (Example.jl or Plots.jl etc?).

pankgeorg:
Just noting that on 

> see how extensions are treated

Extensions are treated badly, as the code for the packages in the distribution are lost, unless they are copied over with `copy_globs=["ext/**"]`.

pankgeorg:
Could we also customize the `versioninfo()` information?

Basically this text:

image

src/PackageCompiler.jl
```diff
@@ -979,6 +999,95 @@ function create_app(package_dir::String,
...
+- `precompile_statements_file`: Extra precompile statements appended to the sysimage build.
+- `incremental::Bool=true`: Whether to extend the current Julia sysimage instead of creating a fresh one.
+- `force::Bool=false`: Overwrite `dist_dir` if it exists.
+- `cpu_target::String=default_app_cpu_target()`: CPU target used when compiling the sysimage.
+- `include_lazy_artifacts::Bool=false`: If `true`, lazy artifacts referenced by dependencies are bundled.
+- `sysimage_build_args::Cmd=```: Additional flags for the Julia process building the sysimage.
+- `include_transitive_dependencies::Bool=true`: If `true`, include transitive dependencies in the sysimage.
+- `include_preferences::Bool=true`: Bundle package preferences into `share/julia/LocalPreferences.toml`.
+- `script::Union{Nothing,String}=nothing`: Optional script executed while generating the sysimage.
+- `copy_globs::Vector{String}=String[]`: Glob patterns for copying package files to the stdlib directory.
```
bauglir: Could we extend this to allow an array of anything that [`glob` accepts as a pattern](https://github.com/vtjnash/Glob.jl?tab=readme-ov-file#usage), i.e. ["strings, `GlobMatch`es, a vector of strings or objects which implement `occursin`, including `RegEx` and `Glob.FilenameMatch` objects"](https://github.com/vtjnash/Glob.jl/blob/d59fbaf37d4f2c124ee614d625d0d6e0e33d749c/README.md?plain=1#L21-L34)?

In particular, I have the case where I need case insensitive matching which as far as I can tell I cannot express using a regular string converted into a `GlobMatch` under the hood. Writing all the possible scenarios out is also a bit cumbersome/noisy.

If the array/`fn` inputs for `glob` would be supported the following would be able to find both files with a single matcher.

```julia-repl
julia> glob("license*", "$(homedir())/.julia/packages/ModelingToolkit/5RGAg/")
String[]

julia> glob([fn"license*"i], "$(homedir())/.julia/packages/ModelingToolkit/5RGAg/")
1-element Vector{String}:
 "~/.julia/packages/ModelingToolkit/5RGAg/LICENSE.md"

julia> glob("license*", "$(homedir())/.julia/packages/OrderedCollections/Xihhq/")
String[]

julia> glob([fn"license*"i], "$(homedir())/.julia/packages/OrderedCollections/Xihhq/")
1-element Vector{String}:
 "~/.julia/packages/OrderedCollections/Xihhq/License.md"
```

ven-k:
`create_distribution  doesn't bundle `include` and `etc` directories.
```
~$ ls .julia/juliaup-depots/juliaup/julia-1.11.9+<custom-distribution>.x64.linux.musl
bin  lib  libexec  local  share
~$ ls .julia/juliaup-depots/juliahub.com/juliaup/juliaup/julia-1.11.9+0.x64.linux.gnu/
bin  etc  include  lib  libexec  LICENSE.md  share
```

ven-k:
TL;DR: Few packages depend on files outside `src` dir. Instead of copying only src and Project.toml, how about we copy all except a few dirs -- which are least likely to impose such dependencies (ex: test, .github)

These packages fail to precompile without following files:

```
julia-distribution-patch/share/julia/stdlib/v1.11/ConstructionBase:
README.md

julia-distribution-patch/share/julia/stdlib/v1.11/DomainSets:
FunctionMaps

julia-distribution-patch/share/julia/stdlib/v1.11/DomainSets/FunctionMaps:
LICENSE  Project.toml  README.md  src  test

julia-distribution-patch/share/julia/stdlib/v1.11/MacroTools:
animals.txt

julia-distribution-patch/share/julia/stdlib/v1.11/OpenSpecFun_jll:
Artifacts.toml
```

This blocks building any sysimage with `incremental=false`

PR #1088, opened 2026-02-09
Widen copy_globs to accept any Glob.glob pattern type

asinghvi17:
## Summary
- Widens `copy_globs` from `Vector{String}` to `Vector` so each element can be anything `Glob.glob` accepts: a `String`, `GlobMatch`, or `Vector` of matchers (including `Regex`, `Glob.FilenameMatch`, etc.)
- Enables case-insensitive matching via e.g. `copy_globs=[[fn"license*"i]]`
- Adds `isa AbstractString` guard on the `**` special-case so non-string patterns fall through to `glob()` directly

Addresses [review feedback from @bauglir](https://github.com/JuliaLang/PackageCompiler.jl/pull/1073#discussion_r2774260236).

## Test plan
- [x] Verified module loads and precompiles cleanly
- [x] Verified `Vector{String}` still works (backward compatible)
- [x] Verified `GlobMatch`, `FilenameMatch`, and mixed vectors are accepted
- [x] Verified `isa AbstractString` guard correctly dispatches string vs non-string patterns

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1021,9 +1021,12 @@ runtime behavior.
 - `include_transitive_dependencies::Bool=true`: If `true`, include transitive dependencies in the sysimage.
 - `include_preferences::Bool=true`: Bundle package preferences into `share/julia/LocalPreferences.toml`.
 - `script::Union{Nothing,String}=nothing`: Optional script executed while generating the sysimage.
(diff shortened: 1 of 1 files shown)


KristofferC:
CI seems to fail

asinghvi17:
This is integrated upstream now so we no longer need this.

PR #1090, opened 2026-02-10
Don't bundle artifacts from full workspace

### LINKED ISSUE #1051: Too many artifacts bundled when using workspaces or separate manifest
I am currently using a `manifest` entry in the Project.toml that I'm building to point to a larger shared dev environment with e.g. PackageCompiler and plotting packages. In 1.12 we'll switch to using workspaces. In both cases I see that e.g. plotting artifacts end up bundled, even though they are not a project dependency.

Specifically this will include the whole workspace/manifest:
https://github.com/JuliaLang/PackageCompiler.jl/blob/57505caad062b5b1985bcbbb2fb15b9b6e9e6d99/src/PackageCompiler.jl#L1492-L1493

Looking at https://github.com/JuliaLang/Pkg.jl/pull/3841 I see that added `load_all_deps_loadable`, which upon some local testing results in the right artifacts being bundled. It is only available in 1.12 though, so we can just wait until 1.12 is supported and then use (always?) use that function if it is defined, updating
https://github.com/JuliaLang/PackageCompiler.jl/blob/57505caad062b5b1985bcbbb2fb15b9b6e9e6d99/src/PackageCompiler.jl#L80-L89
It seems `load_all_deps!` is not defined in any supported Julia versions so we can replace that branch.

We could probably also get it in sooner by copying more Pkg code over, but that may not be worth it.

visr:
Fixes https://github.com/JuliaLang/PackageCompiler.jl/issues/1051

This creates a subproject with only Example.jl, and makes sure that `load_all_deps` only returns Example.jl. This results in `bundle_artifacts` not bundling artifacts from the full workspace.


visr:
This should be good to merge, the new tests are passing.

visr:
Bump

KristofferC:
Thanks!

PR #1096, opened 2026-04-13
use same cpu target as currrent sysimage on 1.13+

KristofferC:


src/PackageCompiler.jl
```diff
@@ -23,7 +23,11 @@ include("library_selection.jl")
 # Arch utils #
 ##############
 
-const NATIVE_CPU_TARGET = "native"
+const DEFAULT_SYSIMAGE_CPU_TARGET = @static if VERSION >= v"1.13-"
+    Sys.sysimage_target()
```
topolarity: Should we use `"sysimage"`? I thought that was a new alias for this in 1.13, IIRC

topolarity: I am a bit confused that this doesn't align with `default_app_cpu_target()` though

KristofferC: > I am a bit confused that this doesn't align with default_app_cpu_target() though

What do you mean? If you mean why app cpu target does not align with "sysimage" it is because you probably want to distribute an app with a wider cpu target than "native" even if you use a locally built julia (that probably defaulted to native)? But maybe you mean something else.

PR #1097, opened 2026-04-22
Avoid error on rebuild with JuliaC

PatrickHaecker:
Force copy of cert.pem in bundle_cert function.

Setting `force=true` avoids aborting JuliaC with an error when building a second time. It should be save to overwrite the cert file, as it has been copied by PackageCompiler.jl and the user requested to regenerate the app.


topolarity (approved):

PatrickHaecker:
@topolarity: Could you please merge the PR? It just hit us again …

topolarity:
Thanks @PatrickHaecker - I'll follow up on https://github.com/JuliaLang/JuliaC.jl/pull/141 to fix this further next week

PR #1099, opened 2026-04-27
Windows: Do not `-Wl,--export-all-symbols` by default

topolarity:
I have been hitting:
```
ld.exe: error: export ordinal too large: 112510
```
in CI for an external project.

This appears to be a failure of https://github.com/JuliaLang/julia/commit/eb4416b16b8a865376da5c76451a4d60516e2c4a. On reflection, that commit probably should not have been merged in the first place. On top of not working that well and causing a severe performance degradation on Windows for image generation, there is a much simpler alternative:

If we drop the `--export-all-symbols` flag in PackageCompiler, Julia already labels any `@ccallable` and required functions in the sysimage as `__declspec(dllexport)` automatically, so the only additional annotation requirements are on any `.c` files provided by the user. Disabling auto-export of all symbols means that the sysimage ends up with ~15 symbols exported, instead of ~45k+

Removing this flag allows us to backport https://github.com/JuliaLang/julia/pull/59736 to Julia 1.10 (fixing the performance regression for package compiler) and avoid the "ordinal too large" error above, at the ~minor cost of the user needing to annotate their code with `dllexport` on Windows. Since explicit exports are the standard platform behavior on Windows, this seems like a reasonable compromise.

**master** (Julia 1.10):
```
C:\Program Files\Microsoft Visual Studio\2022\Community>dumpbin /EXPORTS mylib.dll # before
...
       46195 number of functions
       46195 number of names
```
**PR** (Julia 1.10):
```
C:\Program Files\Microsoft Visual Studio\2022\Community>dumpbin /EXPORTS mylib.dll
...
           6 number of functions
           6 number of names
```

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -806,7 +806,7 @@ function get_extra_linker_flags(version, compat_level, soname)
     soname_arg = soname === nothing ? `` : `-Wl,-soname,$soname`
     rpath_args = rpath_sysimage()
 
-    extra = Sys.iswindows() ? `-Wl,--export-all-symbols` :
+    extra = Sys.iswindows() ? (VERSION >= v"1.11" ? `-Wl,--export-all-symbols` : ``) :
             Sys.isapple() ? `-fPIC $compat_ver_arg $current_ver_arg $rpath_args` :
             Sys.isunix() ? `-fPIC $soname_arg $rpath_args` :
                 error("unknown machine type, not windows, macOS not UNIX")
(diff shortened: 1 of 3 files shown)


KristofferC:
Ironically, CI also fails with

```
C:/Users/runneradmin/AppData/Local/Temp/jl_Cwnmue/artifacts/b17bda08a19173572926f43a48aad5ef3d845e7c/extracted_files/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: error: export ordinal too large: 92817
collect2.exe: error: ld returned 1 exit status
```

topolarity:
Do you have a link? I only see `rm` failures

KristofferC:
Hm, I must have had the logs from https://github.com/JuliaLang/PackageCompiler.jl/pull/1100 up.

PR #1100, opened 2026-04-27
test: Retry `rm` if it failed

topolarity:
Claude spotted this as a potential cause for the frequent Windows CI failures, and I agree it's worth giving it a few retries to have a better chance of success (assuming there's no deeper issue here).

Similar in spirit to https://github.com/JuliaLang/Pkg.jl/blob/c96cfdf70976e8a5cc21fcef53c0ba137f6b2f64/src/utils.jl#L80-L87


KristofferC:
Could be https://github.com/JuliaLang/julia/issues/34700

PR #1106, opened 2026-05-15
ci: declare workflow-level `contents: read` on 1 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 1 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

The following files were left implicit because they reference `GITHUB_TOKEN` / use a write-scope action / trigger on `pull_request_target`. Those scopes are best declared by maintainers: `CompatHelper.yml`, `TagBot.yml`, `ci.yml`.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

### DIFF
--- a/.github/workflows/ci.nightly.yml
+++ b/.github/workflows/ci.nightly.yml
@@ -15,6 +15,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref != 'refs/heads/master' || github.run_number }}
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
 
+permissions:
+  contents: read
+
 jobs:
   test-nightly:
     # We do not run the full test suite on tags, because we already ran it on master before we made the release.


DilumAluthge (approved):

arpitjain099:
Hi maintainers, just a note that the two windows test failures here look pre-existing: `master` itself has the same windows test runs cancelled/failed across the recent 2026-04-28 push runs. The diff in this PR is 3 lines of `permissions: contents: read` on `ci.yml` only, so it shouldn't affect tests. @DilumAluthge has approved.

If there's anything else you'd like me to address, let me know.

PR #1080, opened 2025-12-16
fix some OS checks for FreeBSD

KristofferC:
Should fix #1079. @didier31, can you check?

PR #1109, opened 2026-06-02
Fixes for FreeBSD support

### LINKED ISSUE #1079: [Julia 1.12.2/PackageCompiler] : Unknown os
Hi,

I've installed nicely julia 1.12.2 on FreeBSD 15 via curl -fsSL https://install.julialang.org | sh.

but this command julia --project=. -e 'using PackageCompiler create_app(".", "build/MyApp")' conducts to :

_ERROR: unknown os
Stacktrace:
[1] error(s::String)
@ Base ./error.jl:44
[2] glob_pattern_lib(lib::String)
@ PackageCompiler ~/.julia/packages/PackageCompiler/FE8YQ/src/PackageCompiler.jl:1244
[3] bundle_julia_libraries(dest_dir::String, stdlibs::Vector{String})
@ PackageCompiler ~/.julia/packages/PackageCompiler/FE8YQ/src/PackageCompiler.jl:1305
[4] create_app(package_dir::String, app_dir::String; executables::Nothing, precompile_execution_file::Vector{String}, precompile_statements_file::Vector{String}, incremental::Bool, filter_stdlibs::Bool, force::Bool, c_driver_program::String, cpu_target::String, include_lazy_artifacts::Bool, sysimage_build_args::Cmd, include_transitive_dependencies::Bool, include_preferences::Bool, script::Nothing)
@ PackageCompiler ~/.julia/packages/PackageCompiler/FE8YQ/src/PackageCompiler.jl:877
[5] create_app(package_dir::String, app_dir::String)
@ PackageCompiler ~/.julia/packages/PackageCompiler/FE8YQ/src/PackageCompiler.jl:843
[6] top-level scope
@ none:3
[7] eval(m::Module, e::Any)
@ Core ./boot.jl:489
[8] exec_options(opts::Base.JLOptions)
@ Base ./client.jl:283
[9] start()

ararslan:
The changes are pretty trivial since we can reuse the Linux code paths.

~~I was trying this out locally and it got OOM killed, but it seemed to have been working up to that point. @giordano, you have access to a beefier FreeBSD system than I do, would you be willing to try out this branch?~~ The example from the documentation works as expected.

Fixes #1079. Closes #1080.

### DIFF
--- a/bin/refresh_library_mapping.jl
+++ b/bin/refresh_library_mapping.jl
@@ -136,11 +136,11 @@ end
 function glob_pattern_lib(lib)
     Sys.iswindows() ? lib * ".dll" :
     Sys.isapple() ? lib * "*.dylib" :
-    Sys.islinux() ? lib* "*.so*" :
+    Sys.isunix() ? lib * "*.so*" :
     error("unknown os")
 end
 
-remaining_libs = Set(strip_prefix_suffix.(copy(all_libraries), Sys.isunix() ? "linux" : Sys.isapple() ? "mac" : "windows"))
+remaining_libs = Set(strip_prefix_suffix.(copy(all_libraries), Sys.isapple() ? "mac" : Sys.isunix() ? "linux" : "windows"))
 for (stdlib, data) in d3
     for (os, libs) in data
         for lib in libs
--- a/docs/src/libs.md
+++ b/docs/src/libs.md
@@ -70,7 +70,7 @@ libjulia.dylib
 ...
 ```
 
-(These will have a `.so` extension on Linux, and a `.dll` extension on Windows. There may also
+(These will have a `.so` extension on Linux and BSDs, and a `.dll` extension on Windows. There may also
 be other files in the same directory, depending on your operating system and version of Julia.)
 
 In addition to most of the same keyword arguments as 
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -1318,7 +1318,7 @@ end
 function glob_pattern_lib(lib)
     Sys.iswindows() ? lib * "*.dll" :
     Sys.isapple() ? lib * "*.dylib" :
-    Sys.islinux() ? lib * "*.so*" :
+    Sys.isunix() ? lib * "*.so*" :
     error("unknown os")
 end
 
@@ -1408,7 +1408,7 @@ function bundle_julia_libraries(dest_dir, stdlibs)
 
     # Required libraries
     println("  ├── Base:")
-    os = Sys.islinux() ? "linux" : Sys.isapple() ? "mac" : "windows"
+    os = Sys.isapple() ? "mac" : Sys.iswindows() ? "windows" : "linux"
     for lib in required_libraries[os]
         if Sys.islinux() && lib == "libstdc++"
             matches = libstdcxx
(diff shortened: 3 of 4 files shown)


bin/refresh_library_mapping.jl
```diff
@@ -136,11 +136,11 @@ end
 function glob_pattern_lib(lib)
     Sys.iswindows() ? lib * ".dll" :
     Sys.isapple() ? lib * "*.dylib" :
-    Sys.islinux() ? lib* "*.so*" :
+    Sys.isunix() ? lib * "*.so*" :
     error("unknown os")
 end
 
-remaining_libs = Set(strip_prefix_suffix.(copy(all_libraries), Sys.isunix() ? "linux" : Sys.isapple() ? "mac" : "windows"))
+remaining_libs = Set(strip_prefix_suffix.(copy(all_libraries), Sys.isapple() ? "mac" : Sys.isunix() ? "linux" : "windows"))
```
ararslan, line 143: Note the change of order here. Since `Sys.isunix()` is true on macOS, the `Sys.isapple()` path is never executed, so we had been setting this argument to `"linux"` on macOS. That seemed unintentional, so I changed it (even though no edits to this line were otherwise needed within the scope of the PR) but I can change it back.

ararslan:
Ah, I just saw #1080. @KristofferC, should we close this in favor of that or vice versa?

giordano:
What should I check specifically?

ararslan:
My thought had just been to verify that this package's functionality works as expected, but I was actually able to do that myself (I think—see below). I realized that when I had tried `create_sysimage` earlier and it OOM'd, I was doing a pretty large and complex package. Using Example.jl as in the example in this package's documentation completed successfully. So I think you're off the hook for now, @giordano. 🙂 I appreciate your willingness to help!

<details><summary>A journey of self-discovery (click to expand, if interested)</summary>

While I was able to successfully compile an incremental system image, actually using it as described in the documentation behaved a bit strangely. Granted, this is my first time using PackageCompiler (at least post v1.0), so perhaps this is user error/misunderstanding, and it's also worth noting that I'm using a local build of Julia `master` rather than a tagged version. But when I started a new session using the built system image specified by `-J`, both the InteractiveUtils and REPL stdlibs had to precompile before the banner and prompt were shown. `Base.loaded_modules` showed Example, but `using Example` triggered precompilation of REPLExt, showed a bunch of lines prefixed by `__JL_PRECOMP_VERBOSE_TIMING__`, then said "Package Example not found, but a package named Example is available from a registry" with an installation prompt. Does this indicate an issue with... something that needs fixing?

EDIT: `using Example` actually had to be `using .Example`, which worked fine. I was still surprised by all of the additional compilation though, given that I did not set `incremental=false`. But maybe it's expected since my system image did not include REPL and InteractiveUtils?

EDIT 2: TIL REPL and InteractiveUtils are not in the default sysimage, so perhaps it makes sense that they're not included in the custom built one, thus triggering additional precompilation. So maybe all is well?

EDIT 3: Yeah I think it's fine

</details>

ararslan:
@topolarity, looks like you approved #1080, but that has merge conflicts. Can you take a look at this one? Should be ready to go, and we can close #1080.

topolarity (approved):
Changes look good to me

Thanks @ararslan

ararslan:
Any chance we can tag a new version too, or are there other things you wanted to get in before the next version?

PR #1110, opened 2026-06-02
Separate printing logic from bundle logic

topolarity:
This allows downstream consumers like JuliaC to report sizes more accurately after deleting / stripping libraries or deleting artifacts, etc. or to skip printing these entirely for automated CLI usage.

It's also a nice re-factor anyway.

Drafted by Claude Opus 🤖  and touched up + reviewed by me.

### DIFF
--- a/src/PackageCompiler.jl
+++ b/src/PackageCompiler.jl
@@ -928,7 +928,8 @@ function create_app(package_dir::String,
                     sysimage_build_args::Cmd=``,
                     include_transitive_dependencies::Bool=true,
                     include_preferences::Bool=true,
-                    script::Union{Nothing, String}=nothing)
+                    script::Union{Nothing, String}=nothing,
+                    quiet::Bool=false)
     if filter_stdlibs && incremental
         error("must use `incremental=false` to use `filter_stdlibs=true`")
     end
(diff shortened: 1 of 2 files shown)


src/PackageCompiler.jl
```diff
@@ -1626,45 +1625,112 @@ function bundle_artifacts(ctx, dest_dir; include_lazy_artifacts::Bool)
...
+
+function _print_bundled_artifacts(io::IO, info::BundledArtifacts)
+    # Render only the artifacts that are still present
+    groups = [
+        BundledArtifactGroup(g.package, g.single_jll, filter((a)->ispath(a.path), g.artifacts))
+        for g in info.groups
+    ]
+    filter!(g -> !isempty(g.artifacts), groups)
+    isempty(groups) && return
+    printstyled(io, "PackageCompiler: bundled artifacts:\n")
```
ararslan: Since you aren't specifying any style elements, isn't this the same as `println`?

src/PackageCompiler.jl
```diff
@@ -1626,45 +1625,112 @@ function bundle_artifacts(ctx, dest_dir; include_lazy_artifacts::Bool)
...
+        for (j, a) in enumerate(group.artifacts)
+            if a.shared
+                size_str = "[shared]"
+            else
+                sz = recursive_dir_size(a.path)
+                total_size += sz
+                size_str = pretty_byte_str(sz)
+            end
+            if show_inline
+                println(io, " - ", size_str, "")
```
ararslan: Presumably the trailing empty string is not necessary here?

src/PackageCompiler.jl
topolarity: Both of these quirks are from the old code, but good catch! Will clean up

ararslan: Quirky old code is like some original houses in areas undergoing development, it gives the neighborhood character.

KristofferC:
Is it overkill to have the printer be passed into as a function so that the printing actually happens at the same time things are bundled? I kind of like that experience instead of nothing happening for 10s and then the terminal just got a full blast of stuff.

topolarity:
> Is it overkill to have the printer be passed into as a function so that the printing actually happens at the same time things are bundled? I kind of like that experience instead of nothing happening for 10s and then the terminal just got a full blast of stuff.

Hmm, that's challenging to make work with the JuliaC workflow where we'll strip / remove artifacts downstream so the printing really does have to happen after-the-fact

Any ideas?

topolarity:
@KristofferC I'd like to merge this soon, so that I can use it for some additional JuliaC CI in JuliaLang/julia

Let me know if you object - happy to try a different approach

KristofferC:
Well, my thinking is right now that the printing also works sort of like a progress bar, copying a lot of artifacts usually takes some time and seeing artifact by artifact getting copied visually gives a sense of progress and avoids it looking like PackageCompiler got stuck. Is it possible to keep that "feeling" while still working for the JuliC use case?

topolarity:
> Is it possible to keep that "feeling" while still working for the JuliC use case?

What if we add a proper progress bar? 😉

KristofferC:
Does flipping the axis of progress from vertical to horizontal help JuliaC? :P

topolarity:
> Does flipping the axis of progress from vertical to horizontal help JuliaC? :P

Yes, it does - because then the progress doesn't have to print library / artifact sizes that are incorrect for your final bundle

Just pushed a pair of commits, so let me know what you think. It's not a proper progress bar, but hopefully it's enough to feel OK: `Package Compiler: bundling artifacts (3/8): Cuda_jll`

PR #1115, opened 2026-07-31
Fixes from AI review (gpt) and some CI maintaince

KristofferC:


KristofferC:
71 min on Windows is still kind of unacceptable but it's better at least..

Issue #916, opened 2024-01-15
Cache the "base" sysimage using Scratch.jl


KristofferC:
When creating apps/libraries we first generate a "base" sysimage that does not contain e.g. the precompiled code from the default sysimage that Julia comes bundled with. We could cache that sysimage and reuse it so we do not have to regenerate it over and over.

sloede:
That would be a great feature to have! It slightly relates to some of the suggestions in #858. One question that was hard for me to answer though, is how to ensure that the correct cached base sysimage is determined based on the Julia flags (and other environmental variables?). I am thinking that things like `-O3` or `--check-bounds=yes|no` would probably result in different base images.

Also, given that there is (likely) a myriad of different combinations, do we only want to keep the latest cache image or one for each possible feature set? Also, how to prevent race conditions during generation?

KristofferC:
I think it is fine for code in Base to be compiled with the default arguments. The app/library specific code will then be compiled with whatever specific options are given.

The CPU-target has to be correct though I think.

> Also, how to prevent race conditions during generation?

We could have a PID lock for the sysimage being generated.

sloede:
> I think it is fine for code in Base to be compiled with the default arguments

OK, this makes sense.



> We could have a PID lock for the sysimage being generated.

Sounds good.

In general, if we add such a caching feature, it would be good if we can add functions to also remove the cached file and/or force the regeneration of a new file upon next use.

dhanak:
This seems to be a duplicate of #841. I would very much welcome this feature!

PR #1125, opened 2026-09-14
Speed up builds: cache the fresh base sysimage and overlap the copies

levy:
`create_app` spends almost all of its time in the Julia process that writes the system image. These five commits cut three parts of that, and fix one bug.

**Cache the fresh base system image.** A non-incremental build first builds a base system image from Julia's own sources. It holds nothing from your program, so it is the same file for every build with the same Julia, `cpu_target` and build flags. It was built into a temporary directory and thrown away. It is now cached under a key made from the Julia commit, the version, the target and the flags, so an upgrade gets a new file instead of a stale one. `PACKAGECOMPILER_CACHE_BASE=0` turns the cache off, and `PACKAGECOMPILER_BASE_CACHE` moves it. A build writes into a private directory and renames it into place, so two builds can share the cache. This adds the `SHA` stdlib as a dependency.

**Copy the app while the compiler runs.** `create_app` copied the libraries and the artifacts, and only then started the compiler. Neither waits on the other, so the copies now run on a second task. The Pkg part of the artifact work moved into `collect_bundled_artifacts`, which runs before both tasks, because Pkg is not safe for more than one task. The steps that start a child process of their own now run together, and `PACKAGECOMPILER_JOBS` sets how many.

**Use every CPU for the image threads.** Julia writes the object file of a system image on several threads, but asks for `jl_effective_threads() / 2` of them. On a 32-CPU machine that is 16. The three children that pass `--output-o` now get `JULIA_IMAGE_THREADS` set to all of them. A value that you set yourself is left alone.

**Bug fix.** A failed base system image build left its private directory in the cache.

Where the time goes, measured on `examples/MyApp` on a 32-CPU machine, in an 8-CPU taskset. These are shares, not times: the wall clock of a build moves by two or three times with other load on the machine.

| phase | share |
| --- | --- |
| copy the libraries, the artifacts and the other files | about 2% |
| build the fresh base system image | 0, when the cache answers |
| `Pkg.precompile` on the base system image | about 20% |
| run the precompile execution file | about 8% |
| compile the system image | about 70% |
| link the system image and build the executable | under 1% |

Three settings get confused, and only the first gives a system-image build more cores:

- `JULIA_IMAGE_THREADS` — the threads that write the object file.
- `PACKAGECOMPILER_JOBS` — how many build steps start at once.
- `--threads` on your own Julia process — the threads of PackageCompiler itself. Julia ignores it when it writes an image.

Most of a system-image build can not be made parallel. Its first half infers types under one global lock. One phase used 692 s of CPU in 316 s of wall clock, about two cores, which is a serial front and a parallel tail. More threads shorten the tail only.

How much the overlap is worth: about 2% on `examples/MyApp`. An interleaved A/B/A/B run drifted by more than that, so I would not quote it as a number. It matters more for an app with large artifacts, several precompile execution files, or several executables. The two changes that remove work instead of spreading it are the cache and `cpu_target`. About the target: the default names three processors with `clone_all`, so Julia compiles every function three times. One step of such a build took over six minutes on one core, and the base system image is 225 MB against 165 MB for `cpu_target = "native"`. **This pull request does not change that default.**

Tests: 21 pass, 2 error. The same two error on unmodified master — `create_app` and `create_empty_sysimage`, both in the build of a fresh base system image on Julia 1.13.0. A new test covers the cache key.

**Disclosure:** developed with Claude Code (Opus 5) under my direction. It wrote the code and this text; the measurements were run on my machine. I reviewed the changes. The commits carry an Assisted-by trailer.


KristofferC:
Thanks, caching the base sysimage is a good idea (ref #916).

As a maintainer for a package, though, it is a little bit hard to get a PR such as this one, and I will elaborate on why:

- It uses a non-SOTA model
- It bundles two separate changes (parallelism and caching in one PR)
- It has many auxiliary functions with huge docstrings 
- Comment blocks of multiple lines for a single-line addition
- The PR description is quite large, with a lot of irrelevant information and numbers.

I put up an alternative PR in https://github.com/JuliaLang/PackageCompiler.jl/pull/1126, which I feel is a way to add the caching feature in a more reviewable format.

levy:
I understand your points and I can accept your version.

Should I split this PR to make it more manageable for the rest or would you like to carve out what's valuable there and do it according to your taste? Any help I can provide?

BTW, I'm perfectly fine with closing the PR if the changes go in one way or another. I'd like to hear the reasons if something is rejected though. I'm working towards a fast and reliable incremental compiler which can compile binaries in seconds instead of minutes.

Thank you very much!

KristofferC:
> How much the overlap is worth: about 2% on examples/MyApp. An interleaved A/B/A/B run drifted by more than that, so I would not quote it as a number. It matters more for an app with large artifacts, several precompile execution files, or several executables.

Do you have a concrete example where this is actually significant?

KristofferC:
Regarding

> Use every CPU for the image threads. 

I guess there is a reason julia itself uses only `/ 2` CPUs. PackageCompiler already has a very high memory pressure and I am reluctant to change things from what Julia itself uses. If you want to override this you can use `JULIA_IMAGE_THREADS`.

levy:
> > How much the overlap is worth: about 2% on examples/MyApp. An interleaved A/B/A/B run drifted by more than that, so I would not quote it as a number. It matters more for an app with large artifacts, several precompile execution files, or several executables.
> 
> Do you have a concrete example where this is actually significant?

I do have one but it's not a simple one to send, because it uses incremental compilation and sysimage writing. So you would need a modified Julia version too. This becomes significant when the compilation is down to a few seconds or even below a second. See here:

https://discourse.julialang.org/t/building-an-incremental-sysimage-in-seconds-instead-of-minutes/139352

KristofferC:
> This becomes significant when the compilation is down to a few seconds or even below a second. 

But if compilation is below a second, you could at maximum gain 1 second by overlapping the compilation and the artifact copying? Or am I misunderstanding?

levy:
> > This becomes significant when the compilation is down to a few seconds or even below a second.
> 
> But if compilation is below a second, you could at maximum gain 1 second by overlapping the compilation and the artifact copying? Or am I misunderstanding?

No, you are not. Let's say instead of 2 seconds, the incremental build is down to 1, which is significant.

KristofferC:
> Let's say instead of 2 seconds, the incremental build is down to 1

That is gaining 1 second (2 -> 1)? But that assumes that the copying only takes 1 second, or? The fastest you can go is `max(time to compile, time to copy)`?

levy:
> Regarding
> 
> > Use every CPU for the image threads.
> 
> I guess there is a reason julia itself uses only `/ 2` CPUs. PackageCompiler already has a very high memory pressure and I am reluctant to change things from what Julia itself uses. If you want to override this you can use `JULIA_IMAGE_THREADS`.

You are right.

For the record, on create_sysimg_object_file alone in an 8-CPU lane, A/B/A: 311.6 s at 4 threads, 221.6 s at 8, 282.8 s at 4 again. On a quiet machine, 122 and 124 s against 97 and 119 s. Real, but noisy, and it only shortens the parallel tail.

I dropped it for the PR. The body now carries those numbers under JULIA_IMAGE_THREADS, so anyone with the memory can choose it.

One piece I kept pulls the other way: build_jobs asks jl_effective_threads instead of Sys.CPU_THREADS, so the job count follows the CPU affinity. Under taskset -c 16-23 it is 8, not 32.

I hope it's more acceptable.

levy:
> > Let's say instead of 2 seconds, the incremental build is down to 1
> 
> That is gaining 1 second (2 -> 1)? But that assumes that the copying only takes 1 second, or? The fastest you can go is `max(time to compile, time to copy)`?

Yes, I wasn't clear, that's the whole point. I should have stated that upfront. The reason it was added is because it was measured on a large project.

KristofferC:
But does the copying on a large project really take on the order of a second?

levy:
> But does the copying on a large project really take on the order of a second?

A gigabyte of artifacts plus the 671 MB of Julia libraries takes 1.55 s cold, 0.64 s warm. On examples/MyApp: 1.04 s and 0.14 s. So no, the copying does not take seconds, and the overlap can save about one.

You are right the cache is the change that actually matters. Since your other PR is already merged, I'll save any value that is left in here in another PR.

PR #1133, opened 2026-09-18
Make concurrent uses of the cache not throw on Windows

Drvi:
Our Windows CI builds several binaries using PackageCompiler.jl in parallel and recently it became very flakey with the following stacktrace:
```julia
IOError: open("C:\\...\\scratchspaces\\9b87118b-...\\base_sysimages\\3a471be99561bd85.dll",
              257, 438): resource busy or locked (EBUSY)
  [4] touch(path::String)
  [5] create_fresh_base_sysimage(; cpu_target, sysimage_build_args)
  [7] create_sysimage(...)
  [9] create_library(...)
```

This PR ignores such errors, so the cache updating is best effort only.


KristofferC:
Thanks!

